PR #1, opened 2025-08-14
Prime field modulus and Montgomery constants made static

frozenspider:
Previous implementation of `RandomField` stored `ConfigRef` that described field modulus and precomputed values for Montgomery representation. Moreover, it allowed having `RandomField` with no config at all.
This approach introduced a lot of unnecessary complexity, and wasn't actually useful in any meaningful way.

This PR introduces a static `FieldConfig` that is passed to `RandomField` as a parameter. This also allowed us to get rid of lifetimes in  `RandomField` definition.

The resulting code is shorter, simpler and more performant: as per `RandomFieldArithmetic` benchmark:
```
RandomFieldArithmetic/Multiply/Random128BitFieldElement
                        time:   [3.2569 µs 3.2817 µs 3.3152 µs]
                        change: [−98.302% −98.222% −98.164%] (p = 0.00 < 0.05)

RandomFieldArithmetic/Addition/Random128BitFieldElement
                        time:   [3.2753 µs 3.3093 µs 3.3523 µs]
                        change: [−92.487% −92.357% −92.233%] (p = 0.00 < 0.05)

RandomFieldArithmetic/Division/Random128BitFieldElement
                        time:   [7.2005 ms 7.2498 ms 7.3118 ms]
                        change: [−3.4696% −2.3770% −1.2831%] (p = 0.00 < 0.05)

RandomFieldArithmetic/Negation/Random128BitFieldElement
                        time:   [3.2196 µs 3.2870 µs 3.4107 µs]
                        change: [−79.418% −79.216% −78.937%] (p = 0.00 < 0.05)

RandomFieldArithmetic/Sum/Random128BitFieldElement
                        time:   [167.74 µs 169.54 µs 171.99 µs]
                        change: [−86.156% −86.031% −85.891%] (p = 0.00 < 0.05)

RandomFieldArithmetic/Product/Random128BitFieldElement
                        time:   [1.6908 ms 1.7224 ms 1.7742 ms]
                        change: [−37.438% −36.880% −36.091%] (p = 0.00 < 0.05)
```

### DIFF
--- a/zip-plus/benches/zip_benches.rs
+++ b/zip-plus/benches/zip_benches.rs
@@ -141,6 +141,7 @@ fn open<const P: usize>(group: &mut BenchmarkGroup<WallTime>, spec: usize) {
     let poly = DenseMultilinearExtension::rand(P, &mut rng);
     let (data, _) = BenchZip::commit::<RandomField<FIELD_LIMBS, FC>>(&params, &poly).unwrap();
     let point = vec![1i64; P];
+    let field_point = point.map_to_field();
 
     group.bench_function(
         format!("Open: RandomField<{FIELD_LIMBS}>, poly_size = 2^{P}(Int limbs = {INT_LIMBS}), ZipSpec{spec}, modulus={}", FC::modulus()),
@@ -154,7 +155,7 @@ fn open<const P: usize>(group: &mut BenchmarkGroup<WallTime>, spec: usize) {
                         &params,
                         &poly,
                         &data,
-                        &point.map_to_field(),
+                        &field_point,
                         &mut transcript,
                     )
                     .expect("Failed to make opening");
--- a/zip-plus/benches/zip_benches.rs
+++ b/zip-plus/benches/zip_benches.rs
@@ -1,6 +1,7 @@
 #![allow(non_local_definitions)]
 #![allow(clippy::eq_op)]
 
+use zip_plus::field::config::ConstFieldConfig;
 use std::hint::black_box;
 use ark_std::{
     str::FromStr,
@@ -12,18 +13,7 @@ use criterion::{
 };
 use crypto_bigint::Random;
 use itertools::Itertools;
-use zip_plus::{
-    define_random_field_zip_types,
-    field::{BigInt, ConfigRef, FieldConfig, RandomField},
-    implement_random_field_zip_types,
-    poly_z::mle::{DenseMultilinearExtension, MultilinearExtension},
-    traits::{Config, ConfigReference, FieldMap, ZipTypes},
-    transcript::KeccakTranscript,
-    code::{DefaultLinearCodeSpec, LinearCode},
-    code_raa::RaaCode,
-    pcs::{structs::MultilinearZip, MerkleTree},
-    pcs_transcript::PcsTranscript,
-};
+use zip_plus::{define_random_field_zip_types, field::{BigInt, ConfigRef, FieldConfig, RandomField}, implement_random_field_zip_types, poly_z::mle::{DenseMultilinearExtension, MultilinearExtension}, traits::{Config, ConfigReference, FieldMap, ZipTypes}, transcript::KeccakTranscript, code::{DefaultLinearCodeSpec, LinearCode}, code_raa::RaaCode, pcs::{structs::MultilinearZip, MerkleTree}, pcs_transcript::PcsTranscript, define_field_config};
 
 const INT_LIMBS: usize = 1;
 const FIELD_LIMBS: usize = 4;
@@ -34,6 +24,9 @@ implement_random_field_zip_types!(INT_LIMBS);
 type ZT = RandomFieldZipTypes<INT_LIMBS>;
 type LC = RaaCode<ZT>;
 type BenchZip = MultilinearZip<ZT, LC>;
+
+define_field_config!(Fc, "106319353542452952636349991594949358997917625194731877894581586278529202198383");
+type FC = Fc<FIELD_LIMBS>;
 
 fn encode_rows<const P: usize>(group: &mut BenchmarkGroup<WallTime>, spec: usize) {
     group.bench_function(
@@ -113,7 +106,7 @@ fn commit<const P: usize>(group: &mut BenchmarkGroup<WallTime>, spec: usize) {
                 for _ in 0..iters {
                     let poly = DenseMultilinearExtension::rand(P, &mut rng);
                     let timer = Instant::now();
-                    let _ = BenchZip::commit::<RandomField<FIELD_LIMBS>>(&params, &poly)
+                    let _ = BenchZip::commit::<RandomField<FIELD_LIMBS, FC>>(&params, &poly)
                         .expect("Failed to commit");
                     total_duration += timer.elapsed()
                 }
@@ -124,10 +117,9 @@ fn commit<const P: usize>(group: &mut BenchmarkGroup<WallTime>, spec: usize) {
     );
 }
 
-fn open<const P: usize>(group: &mut BenchmarkGroup<WallTime>, modulus: &str, spec: usize) {
+fn open<const P: usize>(group: &mut BenchmarkGroup<WallTime>, spec: usize) {
     let mut rng = test_rng();
-    let config = FieldConfig::new(BigInt::<FIELD_LIMBS>::from_str(modulus).unwrap());
-    let field_config = ConfigRef::from(&config);
+    let field_config = ConfigRef::from(&FC::FIELD_CONFIG);
 
     type T = KeccakTranscript;
     let mut keccak_transcript = T::new();
@@ -136,16 +128,16 @@ fn open<const P: usize>(group: &mut BenchmarkGroup<WallTime>, modulus: &str, spe
     let params = BenchZip::setup(poly_size, linear_code);
 
     let poly = DenseMultilinearExtension::rand(P, &mut rng);
-    let (data, _) = BenchZip::commit::<RandomField<FIELD_LIMBS>>(&params, &poly).unwrap();
+    let (data, _) = BenchZip::commit::<RandomField<FIELD_LIMBS, FC>>(&params, &poly).unwrap();
     let point = vec![1i64; P];
 
     group.bench_function(
-        format!("Open: RandomField<{FIELD_LIMBS}>, poly_size = 2^{P}(Int limbs = {INT_LIMBS}), ZipSpec{spec}, modulus={modulus}"),
+        format!("Open: RandomField<{FIELD_LIMBS}>, poly_size = 2^{P}(Int limbs = {INT_LIMBS}), ZipSpec{spec}, modulus={}", FC::MODULUS),
         |b| {
             b.iter_custom(|iters| {
                 let mut total_duration = Duration::ZERO;
                 for _ in 0..iters {
-                    let mut transcript = PcsTranscript::<RandomField<FIELD_LIMBS>>::new();
+                    let mut transcript = PcsTranscript::<RandomField<FIELD_LIMBS, FC>>::new();
                     let timer = Instant::now();
                     BenchZip::open(
                         &params,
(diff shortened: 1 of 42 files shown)


cupicmarko:
I talked about similar changes few months ago with Isaac Villalobos and Matthew Klein, I suggested similar approach, but they mentioned that similar libraries exist like PrimeFields from arkworks, they wanted to keep config mutable and non-static because static doesn't allow for any interoperability between configs in runtime.

frozenspider:
> I talked about similar changes few months ago with Isaac Villalobos and Matthew Klein, I suggested similar approach, but they mentioned that similar libraries exist like PrimeFields from arkworks, they wanted to keep config mutable and non-static because static doesn't allow for any interoperability between configs in runtime.

That's what I was striving to, to eventually unite it with some existing library through a thin wrapper. It doesn't seem we're using config interop in a meaningful way - and if we ever want it, we can always convert a value from one field to another though demontgomery -> montgomery (as it was done anyways)

zip-plus/src/code_raa.rs
```diff
@@ -316,10 +320,10 @@ mod tests {
...
         struct MismatchedZipTypes;
         impl ZipTypes for MismatchedZipTypes {
-            type N = Int<1>; // 64 bits
-            type L = Int<2>; // 128 bits
-            type K = Int<1>; // 64 bits - INSUFFICIENT
-            type M = Int<4>; // 256 bits
+            type K = Int<1>; // 128 bits
+            type L = Int<2>; // 64 bits
+            type M = Int<4>; // 64 bits - INSUFFICIENT
+            type N = Int<1>; // 256 bits
```
NiDimi: I think when you changed the letters the comment no longer make sense

zip-plus/benches/zip_benches.rs
```diff
@@ -136,23 +139,22 @@ fn open<const P: usize>(group: &mut BenchmarkGroup<WallTime>, modulus: &str, spe
...
-                    let mut transcript = PcsTranscript::<RandomField<FIELD_LIMBS>>::new();
+                    let mut transcript = PcsTranscript::<RandomField<FIELD_LIMBS, FC>>::new();
                     let timer = Instant::now();
                     BenchZip::open(
                         &params,
                         &poly,
                         &data,
-                        &point.map_to_field(field_config),
-                        field_config,
+                        &point.map_to_field(),
```
NiDimi: Lets move the `map_to_field` outside teh bench

zip-plus/src/conversion.rs
```diff
@@ -1,23 +1,16 @@
...
-use crate::traits::{
-    BigInteger, Config, ConfigReference, Field, FieldMap, Integer, PrimitiveConversion, Uinteger,
-    Words,
+use crate::{
+    field::config::{FieldConfigBase, FieldConfigOps},
+    traits::{BigInteger, Field, FieldMap, Integer, PrimitiveConversion, Uinteger, Words},
 };
+use ark_std::{Zero, vec::Vec};
 
 // Implementation of FieldMap for signed integers
```
NiDimi: Nit pick but we use for both signed and unsigned now

zip-plus/src/conversion.rs
```diff
@@ -104,228 +93,178 @@ where
...
+        let result = RandomField::<1, _>::from_bytes_be(&bytes);
+        assert_eq!(result, Some(expected.clone()));
     }
 
     #[test]
     fn converts_from_bytes_le_with_config_out_of_range() {
-        let config = field_config!(23);
-        let config = ConfigRef::from(&config);
-
         let bytes = [0x65, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00]; // Value: 101 (modulus is 23)
```
NiDimi: Actually not 101

zip-plus/src/conversion.rs
```diff
@@ -606,13 +515,9 @@ mod bigint_field_map_tests {
 
     #[test]
     fn test_bigint_reference() {
```
NiDimi: This looks like it meant to compare owned vs. reference mapping but i think both sides map the owned value now

zip-plus/src/field.rs
```diff
@@ -258,129 +117,73 @@ impl<'cfg, const N: usize> Field for RandomField<'cfg, N> {
...
-impl<const N: usize> UniformRand for RandomField<'_, N> {
+impl<const N: usize, FC: FieldConfig<BigInt<N>>> UniformRand for RandomField<N, FC> {
     fn rand<R: ark_std::rand::Rng + ?Sized>(rng: &mut R) -> Self {
         let value = BigInt::rand(rng);
 
-        Self { value, config: None }
+        Self {
+            value,
+            phantom_data: PhantomData,
+        }
```
NiDimi: Use `<Self as Field>::rand(rng)` to store in montgomery form

zip-plus/src/field.rs
```diff
@@ -258,129 +117,73 @@ impl<'cfg, const N: usize> Field for RandomField<'cfg, N> {
...
-impl<const N: usize> Random for RandomField<'_, N> {
+impl<const N: usize, FC: FieldConfig<BigInt<N>>> Random for RandomField<N, FC> {
     fn random(rng: &mut (impl ark_std::rand::RngCore + ?Sized)) -> Self {
         let value = BigInt::rand(rng);
 
-        Self { value, config: None }
+        Self {
+            value,
+            phantom_data: PhantomData,
+        }
```
NiDimi: `Use <Self as Field>::rand(rng)` to store in montgomery form

zip-plus/src/field.rs
```diff
@@ -17,151 +18,45 @@ pub use biginteger::{
...
-    pub fn from_i64(value: i64, config: ConfigRef<'cfg, N>) -> Option<RandomField<'cfg, N>> {
-        let config_ref = match config.reference() {
-            Some(config) => config,
-            None => {
-                panic!("Cannot convert signed integer to prime field element without a modulus")
-            }
-        };
-
-        if BigInt::from(value.unsigned_abs()) >= *config_ref.modulus() {
+    pub fn from_i64(value: i64) -> Option<RandomField<N, FC>> {
```
NiDimi: all the `fromu63,u32,u16,u8` reduce modulo p but `from_i64` just returns None. I think it would be better if we had consistent policy

zip-plus/src/field/arithmetic.rs
```diff
@@ -72,107 +77,63 @@ impl<const N: usize> Add<u32> for RandomField<'_, N> {
...
-        // self.with_aligned_config_mut(
-        //     rhs.into(),
-        //     |lhs, rhs, config| {
-        //         config.add_assign(lhs, rhs);
-        //     },
-        //     |lhs, rhs| {
-        //         panic!();
-        //     },
-        // );
+        FC::add_assign(&mut self.value, &rhs.into());
```
NiDimi: You are mixing montgomery and non-montigomery. into will turn into a bigInt try this:

`let rhs_f: RandomField<N, FC> = rhs.map_to_field();
        FC::add_assign(&mut self.value, &rhs_f.value);`

zip-plus/src/field/arithmetic.rs
```diff
@@ -43,27 +46,29 @@ macro_rules! impl_ops {
...
 }
 
-impl_ops!(impl('cfg, const N: usize) for RandomField<'cfg, N>, Add, add, AddAssign, add_assign);
-impl_ops!(impl('cfg, const N: usize) for RandomField<'cfg, N>, Sub, sub, SubAssign, sub_assign);
-impl_ops!(impl('cfg, const N: usize) for RandomField<'cfg, N>, Mul, mul, MulAssign, mul_assign);
-impl_ops!(impl('cfg, const N: usize) for RandomField<'cfg, N>, Div, div, DivAssign, div_assign);
+impl_ops!(impl('cfg, const N: usize, FC: FieldConfig<BigInt<N>>) for RandomField<N, FC>, Add, add, AddAssign, add_assign);
+impl_ops!(impl('cfg, const N: usize, FC: FieldConfig<BigInt<N>>) for RandomField<N, FC>, Sub, sub, SubAssign, sub_assign);
+impl_ops!(impl('cfg, const N: usize, FC: FieldConfig<BigInt<N>>) for RandomField<N, FC>, Mul, mul, MulAssign, mul_assign);
+impl_ops!(impl('cfg, const N: usize, FC: FieldConfig<BigInt<N>>) for RandomField<N, FC>, Div, div, DivAssign, div_assign);
```
NiDimi: I might be missing something, but are the lifetimes used anywhere?

zip-plus/src/field/config.rs
```diff
@@ -1,8 +1,5 @@
+use crate::field::BigInt;
 use num_traits::Zero;
-use crate::{
-    field::BigInt,
-    traits::{Config, ConfigReference},
-};
 
 #[macro_export]
 macro_rules! mac_with_carry {
```
NiDimi, line 5: A bit outside teh code but we have both a macro and fn with the same name. It would be better if we could change that

zip-plus/src/field/config.rs
```diff
@@ -27,56 +24,132 @@ pub fn mac_with_carry(a: u64, b: u64, c: u64, carry: &mut u64) -> u64 {
...
-        self.reduce_modulus(a, c);
+        Self::reduce_modulus(a, c);
     }
 
-    pub fn sub_assign(&self, a: &mut BigInt<N>, b: &BigInt<N>) {
+    fn sub_assign(a: &mut BigInt<N>, b: &BigInt<N>) {
         // If `other` is larger than `self`, add the modulus to self first.
         if b > a {
-            a.add_with_carry(&self.modulus);
+            a.add_with_carry(&Self::modulus());
```
NiDimi, line 139:
```suggestion
           // Add modulus first so the subtraction doesn't underflow; carry is intentionally
           // ignored (we operate modulo 2^(64*N) here and the subsequent subtraction pulls it back).
            a.add_with_carry(&Self::modulus());
```

NiDimi (changes requested):
Left some comments. Some of them are outside the scope but spotted them as I was going through the PR. Really cool work

zip-plus/src/field/arithmetic.rs
frozenspider: You're right, those are leftovers, thanks

zip-plus/src/field/config.rs
frozenspider: I expect we get rid of them altogether soon enough

zip-plus/src/code_raa.rs
frozenspider: Damn, that's why `reorder_impl_items` of `rustfmt` is unstable :( Disabled it

zip-plus/src/conversion.rs
frozenspider: Tbh that comment was useless anyway

zip-plus/src/conversion.rs
frozenspider: It is 101 though, we're reading them in LE order and 0x65 is 101

zip-plus/src/conversion.rs
frozenspider: Previous version of this test was also not doing anything meaningful - looks like an AI slop to me. Removed the test.

zip-plus/src/field.rs
frozenspider: We actually don't need this at all - it pretty much duplicates what `map_to_field` does, and is a leftover

zip-plus/src/field.rs
frozenspider: Removed excessive implementations of random generation

zip-plus/src/conversion.rs
frozenspider: Anyway, removed both the test and methods `from_xxx_bytes` - they weren't used outside of tests

frozenspider:
Thanks @NiDimi for the thorough review and nice catches!

NiDimi (approved):
Latest changes look good. Nice that the CI also got fixed :+1:

PR #2, opened 2025-09-19
Use abstract rings for Zip and linear code

frozenspider:
Instead of `N, L, K, M` usize constant bounds for `Int`s, both Zip and linear code now operates on 4 abstract rings:
* `N => EvaluationRing` 
* `L` (previously: element width of the encoding matrix of a linear code) is no longer a generic, instead it's made into a property of linear code. `PrimeField` elements must be obtainable from it, but otherwise code is free to place any constraints on the rings that it needs to make it work.
* `K => CodewordRing`
* Added new ring `ChallengeRing`, used to generate challenges for random linear combination of codewords. Previously it was the same as evaluation ring.
* `M => LinearCombinationRing`

These ring aliases are implemented by `Int<_>` rings, so not much changes from a user perspective, except you have to specify `ChallengeRing` separately.
 
## Other major changes
* Added a new trait `MulByScalar` to represent an ability to multiply a given ring element by `ChallengeRing` (and field elements with each other, as they are being combined too)
* `LinearCode` no longer has generic `encode_wide`, instead 3 encode methods are now defined as follows:
  * `encode: EvaluationRing => CodewordRing`
  * `encode_wide: LinearCombinationRing => LinearCombinationRing`
  * `encode_f: PrimeField => PrimeField`
* Tweaked `RaaCode` to have a configurable `check_for_overflows`. This is needed to keep encoding times on par with original Zinc encoding times - using `checked_add` adds ~20% overhead.
* All to- and from-bytes encoding is now done through a new trait `Transcribable`. It can use whatever byte order as long as it allows serialization and deserialization cross-platform. (Actually uses little-endian order as being more efficient and more natural when applied to underlying `crypto-bigint` types)
  * This allows to make transcription reads/writes much more generalized, removing a bunch of type-specific methods
  * This also made `AsWords` trait unnecessary, removed it
  * Cross-platform deserialization wasn't actually tested yet, left a TODO

## Additional changes
* `ConstRing` is now auto-implemented
* `dense` multilinear polynomial module moved from `poly` to `poly::mle` to free up module name
* `num_proximity_testing` no longer depends on `log2_q`. It was unused previously anyway, and if we need it in future, we'd need to determine what `log2_q` is for an arbitrary ring.

## Performance
(single-threaded)
* Encoding and MT construction is on par with original Zinc.
* Commitment overall is ~35% faster.
* Open and verify is on par with https://github.com/NethermindEth/zinc/pull/142 (which is much faster than original Zinc)


zip-plus/src/code/raa.rs
```diff
@@ -169,6 +202,25 @@ where
     }
 }
 
+#[allow(clippy::arithmetic_side_effects)]
+fn accumulate_unchecked<I>(input: &mut [I])
```
cupicmarko, line 215: Based on the name, I would suggest marking the function unsafe and removing unsafe block inside.

zip-plus/src/traits.rs
```diff
@@ -1,37 +1,49 @@
...
+    /// Creates a new instance from a byte buffer.
+    /// The buffer must be exactly `NUM_BYTES` long.
+    fn from_transcription_bytes(bytes: &[u8]) -> Self;
 
-pub trait FromBytes {
-    fn from_be_bytes(bytes: &[u8]) -> Self;
-    fn from_le_bytes(bytes: &[u8]) -> Self;
+    /// Transcribes the current instance into a byte buffer.
+    /// Buffer must be exactly `NUM_BYTES` long.
+    fn to_transcription_bytes(&self, buf: &mut [u8]);
```
cupicmarko: I like this, it doesn't make any heap allocation, I would only suggest renaming it into something like `write_transcription_bytes`, `to_` generally implies that a conversion returns a new, owned value.

cupicmarko (commented):
I didn't finish the review yet, but I'll mark the PR approved once I finish, I don't see any changes to request or any questions. If performance are the same or better, this is a nicer, cleaner code.

I would reconsider `#[inline(always)]` flags, because they can increase the resulting code size, which is usually fine, but it can affect performance on CPUs with small instruction cache (Cortex-M can have 8-16KB for example), so I would suggest using `#[inline]` and letting the compiler decide on inlining. I don't know how real this scenario is, just mentioning it.

zip-plus/src/traits.rs
frozenspider: Good point!

zip-plus/src/code/raa.rs
frozenspider: This function is not unsafe in the usual Rust sense (no undefined behaviour). I agree that it makes sense to mark function unsafe for internal reasons too, but in this case this intention is hard to propagate. Making it `unsafe` will just make `encode_inner` to use `unsafe` call wrapper twice - however, both it and `accumulate_unchecked` are internal functions not visible to the outside world, so it doesn't feel like we're accomplishing much.
On the other hand, `unsafe` block inside this function is unsafe in memory manipulation sense, so it makes sense to keep `unsafe {}` block wrapper there.
Overall, I think in this case I'd prefer to keep it the way it is now.

frozenspider:
Thanks @cupicmarko!
As for `#[inline(always)]`, I wanted to use it coupled with `#[repr(transparent)]` to give compiler more a strong hint to remove the wrapper altogether - calls should just be forwarded to the underlying repr. This might not be needed though.

zip-plus/src/code/raa.rs
```diff
@@ -45,14 +57,16 @@ impl<const N: usize, const L: usize, const K: usize, const M: usize> RaaCode<N,
...
-        let num_proximity_testing = spec.num_proximity_testing(log2_q, row_len, n_0);
+        let num_proximity_testing = spec.num_proximity_testing(row_len, n_0);
 
         // Width of each entry in codeword vector, in bits.
         // For RAA it's initial_bits + 2*log(repetition_factor) + num_variables
         let codeword_width_bits = {
-            let initial_bits = crypto_bigint::Int::<N>::BITS;
+            // let initial_bits = crypto_bigint::Int::<N>::BITS;
+
+            let initial_bits = size_of::<Eval>() as u32 * 8;
```
maksimryndin: can we use here a safer conversion with `[try_from](https://doc.rust-lang.org/stable/std/primitive.u32.html#impl-TryFrom%3Cusize%3E-for-u32)`?

zip-plus/src/code/raa.rs
```diff
@@ -169,6 +202,25 @@ where
...
 
+#[allow(clippy::arithmetic_side_effects)]
+fn accumulate_unchecked<I>(input: &mut [I])
+where
+    I: Zero + for<'a> AddAssign<&'a I> + Clone,
+{
+    for i in 1..input.len() {
+        // This allows us to circumvent Rust bounds checking
+        let input_i: &mut I = unsafe {
+            let input_i: *mut I = &mut input[i];
```
maksimryndin: So, miri doesn't complain for this?
Could we use https://doc.rust-lang.org/stable/std/primitive.slice.html#method.get_unchecked?

zip-plus/src/pcs/structs.rs
```diff
@@ -1,79 +1,91 @@
...
+    Eval: EvaluationRing,
+    Cw: CodewordRing,
+    Chal: ChallengeRing,
+    Comb: LinearCombinationRing<Eval, Cw, Chal>,
+    C: LinearCode<Eval, Cw, Comb>,
+{
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn setup(poly_size: usize, linear_code: C) -> MultilinearZipParams<Eval, Cw, Comb, C> {
+        assert!(poly_size.is_power_of_two());
+        let num_vars = poly_size.ilog2() as usize;
```
maksimryndin, line 33: try_from with expect?

zip-plus/src/pcs/tests.rs
```diff
@@ -5,44 +5,19 @@
...
-            if columns.insert(i) {
-                inserted += 1;
-                if inserted == count {
-                    break;
-                }
-            }
-        }
-        inserted
+        let mut bytes = vec![0u8; T::NUM_BYTES];
+        let counter_bytes = self.counter.to_be_bytes();
```
maksimryndin: just for double check - in the PR description you mention "little-endian order"

maksimryndin (commented):
Awesome work, @frozenspider ! 👍 

Probably, in future optimizations it is worth also to try an alternative to crypto-bigint

https://github.com/recmo/uint by Remco et al

zip-plus/src/code/raa.rs
frozenspider: Does it really make sense to use it though? I don't think it's reasonable to expect `Eval` to be over 500 MB in size

zip-plus/src/pcs/structs.rs
frozenspider: Since `poly_size` itself is `usize`, `poly_size.ilog2()` is guaranteed to fit `usize`

zip-plus/src/pcs/tests.rs
frozenspider: Thanks! This is test code, but it makes sense to keep everything LE for consistency

zip-plus/src/code/raa.rs
maksimryndin: Sometimes we are wrong in our assumptions :)
Do you think it will take you more time to fix it rather than writing a comment? :)

zip-plus/src/code/raa.rs
frozenspider: Yeah, this is not UB.
Used `get_unchecked` for index access.

maksimryndin: Another variant can be (godbolt generates similar code but I haven't benchmark) - just of curiosity =)

```rust
let mut input: Vec<u32> = vec![0, 1, 2, 3, 4];
 for i in 1..input.len() {
      unsafe {
          let (left, right) = input.split_at_mut_unchecked(i);
          *right.get_unchecked_mut(0) += left.get_unchecked(i-1);
      }
  }
```

zip-plus/src/code/raa.rs
cupicmarko: I see your point, can we add debug_asserts that check for the UB?

cupicmarko:
> Thanks @cupicmarko! As for `#[inline(always)]`, I wanted to use it coupled with `#[repr(transparent)]` to give compiler more a strong hint to remove the wrapper altogether - calls should just be forwarded to the underlying repr. This might not be needed though.

That seems good, although, I would assume that the compiler is already doing it on its own. It's not something that is needed to be changed, just NIT.

cupicmarko (approved):
Looks good! Just some NIT comments, nothing else

zip-plus/src/code/raa.rs
frozenspider: This code looks a little nicer, but it consistently gives ~5% slower encoding time in `Zip/EncodeRows` sadly

zip-plus/src/code/raa.rs
frozenspider: Asked a follow-up question on Slack, let's do it in the next PR

PR #3, opened 2025-09-22
Allow using primitive signed types as rings in Zip

frozenspider:
* Minimized direct usage of `crypto_bigint::Random`, replacing its usage with more general `rand::Distribution` instead.
* Added `FromRef<T>` trait analogous to `From<&T>` - mainly to allow us to implement it for primitive types.
* Implemented `AsPackable` and `Transcribable` for primitive rings, and added `MulByScalar<primitive>` variant to `Int<_>` 

All of this allows us to use primitive signed types (mainly `i32`, `i64` and `i128`) as rings in Zip instead of `Int<_>`.
(According to my local tests, using `i64`, `i128` instead of `Int<1>`, `Int<2>` for encoding leads to ~30% performance gain)


crypto-primitives/src/ring/crypto_bigint_int.rs
```diff
@@ -394,13 +413,20 @@ impl<const LIMBS: usize> IntRing for Int<LIMBS> {}
...
+impl<const LIMBS: usize> Distribution<Int<LIMBS>> for StandardUniform {
+    fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> Int<LIMBS> {
+        crypto_bigint::Random::random(rng)
+    }
+}
+
+#[cfg(feature = "rand")]
 impl<const LIMBS: usize> crypto_bigint::Random for Int<LIMBS> {
-    fn random<R: rand_core::RngCore + ?Sized>(rng: &mut R) -> Self {
+    fn random<R: RngCore + ?Sized>(rng: &mut R) -> Self {
```
maksimryndin, line 425: should we introduce a trait bound for CSPRNG? (to make it tighter?)
https://docs.rs/rand/latest/rand/trait.CryptoRng.html

https://docs.rs/crypto-bigint/latest/crypto_bigint/prelude/trait.Random.html

frozenspider: I don't think this is a good idea - if we need a CSPRNG, we can still supply it here, but often we do not (like in tests or benchmarks), and I don't feel like restricting it would be super useful - after all, we can't prevent using non-CSPRNG for primitive types anyway.

maksimryndin: The type system should express the intentions and invariants
For the tests you should just implement a CSPRNG mock trait.

For benchmarks I am surprised that we don't use the same CSPRNG as in production - it seems a faulty approach?

> we can't prevent using non-CSPRNG for primitive types anyway.

could you please elaborate?

frozenspider: > The type system should express the intentions and invariants

Right, it just doesn't make too much sense to me to deny non-crypto RNG, as those have perfectly valid use cases.

> For benchmarks I am surprised that we don't use the same CSPRNG as in production - it seems a faulty approach?

These are used to generate random polynomials and messages as inputs for PCS and linear code. We don't need them to be cryptographically secure.

> could you please elaborate?

Even if we restrict our types to only accept CSPRNG, user can always do
```rust
let mut rng = todo!("Non-secure RNG");
let x: u64 = rng.random();
```

Either way a canonical way to implement randomness compatible with `rand` crate (the line we're discussing is for non-standard `crypto_bigint::Random`) is via `impl Distribution<X> for StandardUniform`, and it has a predefined signature.

maksimryndin: >Right, it just doesn't make too much sense to me to deny non-crypto RNG, as those have perfectly valid use cases.

Let's just ensure that in places where the CSPRNG is expected, the trait bound enforces that?

> Even if we restrict our types to only accept CSPRNG, user can always do

Yeah, I understand such ways but they're not limited to primitives - we can do it for any type :)

frozenspider: > Let's just ensure that in places where the CSPRNG is expected, the trait bound enforces that?

We could technically introduce our own trait for that, but avoiding the question of whether or not it would be a good idea do it, it's definitely out of scope for this PR

maksimryndin: then let's create an issue for that, @frozenspider ? :)

frozenspider: Done 🙂 #5

cupicmarko (approved):
I don't have any comment on the code specifics as it seems good. I have a suggestion of adding blanket implementation of FromRef for T. Otherwise, if benches and tests are good, this looks great!

frozenspider:
Thanks for the review @cupicmarko! Yeah, I wanted to implement blanket `FromRef`, either
`impl<'a, T, S> FromRef<S> for T where T: From<&'a S>`
or at least
`impl<T> FromRef<T> for T where T: Clone`
Sadly, both of them conflict with
`impl<const LIMBS: usize, const LIMBS2: usize> FromRef<Int<LIMBS2>> for Int<LIMBS>`
(as there's no way to restrict `LIMBS != LIMBS2`) and many other similar impls, and that defeats the main purpose of this trait 😞

PR #4, opened 2025-10-01
Zip+

frozenspider:
Implement Zip+ - a variant of Zip that works over Z[X] instead of Z.

Main logic changes per stage:
* **Commit**: (unchanged)
* **Test**: In addition to coefficient scalars for linear combination, verifier now also supplies alphas - scalars to be used as a point at which to evaluate each polynomial coefficient at, before doing linear combination.
* **Evaluate**: Verifier now provides a projecting element that's used to make canonical projections of polynomials onto a field. Also, evaluate made to return an evaluation (mapped to a field).

Main code changes:
* Zip-specific types (mostly rings) are now grouped into `ZipTypes` (as they once were).
* Added a new dense univariate polynomial with straightforward implementation. Polynomial can be evaluated at a given point (each coordinate being interpreted as a `X^i`).
* Added `ProjectableToField` trait that prepares canonical projection of a polynomial onto a finite field using a given field element. This element is not used when projecting zero-degree polynomials.
* A bunch of types renamed: `MultilinearZip -> ZipPlus`, `MultilinearZipData` -> `ZipPlusHint`, etc.
* `zip_benches` has been split onto two benchmarks: `zip_benches` and `zip_plus_benches`, with common code in `zip_common`
* Fixed a major design issue in `combine_rows` leading to large chunks of data being copied over and over, this leads to ~10x performance gain in `open`


zip-plus/src/poly/dense.rs
```diff
@@ -0,0 +1,468 @@
...
+
+impl<R: Ring, const DEGREE: usize> Zero for DensePolynomial<R, DEGREE> {
+    fn zero() -> Self {
+        Self {
+            coeff_0: R::zero(),
+            coeffs: array::from_fn::<_, DEGREE, _>(|_| R::zero()),
+        }
+    }
+
+    fn is_zero(&self) -> bool {
```
cupicmarko, line 131: If I understand the struct correctly, should the coeff_0 be also a zero?

zip-plus/src/field/conversion.rs
```diff
@@ -111,13 +138,44 @@ where
...
+    R: Ring + Transcribable,
+    ConstMontyField<Mod, LIMBS>: FromRef<R>,
 {
-    fn from_ref(value: &T) -> Self {
-        Self::from(value.clone())
+    fn prepare_projection(
+        sampled_value: &ConstMontyField<Mod, LIMBS>,
+    ) -> impl Fn(&Self) -> ConstMontyField<Mod, LIMBS> + 'static {
+        let degree_bound: u32 = DEGREE.try_into().expect("Degree bound must fit into u32");
+        let r_powers = (1..=degree_bound)
```
cupicmarko: We can store previous power to get the next next, this quadratic but can be linear.

cupicmarko (approved):
After fixing the is_zero (if it needs fixing it), and performance is not affected, this looks good

zip-plus/src/poly/dense.rs
frozenspider: Good catch, thanks!

zip-plus/src/field/conversion.rs
frozenspider: Makes sense, thanks!

frozenspider:
Thanks for the review! Yes, the performance in my tests didn't regress, while `open` performance greatly improved

PR #6, opened 2025-10-09
Separate linear code and adjust benchmark parameters

frozenspider:
* Separated `LinearCode` from `ZipTypes` to make it dependent on `ZipTypes`. This simplifies linear code typing.
* Fixed an issue with `RaaCode` incorrectly calculating codeword bit-width upper bound for polynomials (it used polynomial size rather than the size of its coefficients)
* Adjusted benchmarks parameters as per discussion with @albert-garreta. Used primitive types instead of `Int<_>` where it was applicable.
* Removed `LinearCodeSpec`, replaced it with constants `ZipTypes::NUM_COLUMN_OPENINGS` and `LinearCode::REPETITION_FACTOR`
* Added new bounds check that ensures that bit-width of the linear combinations ring is enough to fit linear combinations

### DIFF
--- a/zip-plus/benches/zip_benches.rs
+++ b/zip-plus/benches/zip_benches.rs
@@ -13,13 +13,13 @@ const INT_LIMBS: usize = WORD_FACTOR;
 
 struct BenchZipTypes {}
 impl ZipTypes for BenchZipTypes {
-    type EvalR = Int<{ INT_LIMBS }>;
+    type EvalR = i32;
     type Eval = Self::EvalR;
-    type CwR = Int<{ INT_LIMBS * 4 }>;
+    type CwR = i64;
     type Cw = Self::CwR;
-    type Chal = Int<{ INT_LIMBS }>;
-    type Pt = Int<{ INT_LIMBS }>;
-    type CombR = Int<{ INT_LIMBS * 8 }>;
+    type Chal = i128;
+    type Pt = i128;
+    type CombR = Int<{ INT_LIMBS * 4 }>;
     type Comb = Self::CombR;
     type Code = RaaCode<Self::Eval, Self::Cw, Self::Comb>;
 }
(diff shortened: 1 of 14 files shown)


zip-plus/benches/zip_benches.rs
```diff
@@ -13,20 +13,21 @@ const INT_LIMBS: usize = WORD_FACTOR;
 
 struct BenchZipTypes {}
 impl ZipTypes for BenchZipTypes {
-    type EvalR = Int<{ INT_LIMBS }>;
+    const NUM_COLUMN_OPENINGS: usize = 650;
```
maksimryndin, line 16: How is this number obtained? Probably, it is worth to add a comment?

zip-plus/src/code/raa.rs
```diff
@@ -374,21 +332,24 @@ mod tests {
...
-        test_raa::<Int<N>, Int<K>, Int<M>, _>(16, |code| {
+        test_raa::<TestZipTypes<N, K, M>, _>(16, |code| {
             let a: Vec<Int<N>> = (1..=4).map(Int::<N>::from).collect();
 
             let encode_a: Vec<Int<K>> = code.encode(&a);
-
             assert_eq!(
                 encode_a,
-                [0x11, 0x1C, 0x21, 0x28, 0x3C, 0x48, 0x49, 0x58].map(Int::<K>::from)
+                [
```
maksimryndin, line 341: How does this known output come from?

maksimryndin (approved):
LGTM

zip-plus/benches/zip_benches.rs
frozenspider: This is a number picked by @albert-garreta, not sure if there's a certain formula behind that

zip-plus/src/code/raa.rs
frozenspider: Since seeds we're picking in tests are provided by a predictable mock transcript (N.B.: in further PRs, seeds become hardcoded), we're just freezing the sampled output and making sure it doesn't change without a clear reason.

PR #7, opened 2025-10-09
Separate testing and evaluation phases

frozenspider:
Previously, `test` and `evaluate` PCS phases were combined in a single `open` phase. This made it impossible to use/benchmark them separately, and also cumbersome to work with.
* All phases renamed to have `phase_` prefix (could've been made into submodule but I think it makes a structure a bit too deep). Code itself is unchanged.
  * Phases no longer accept an untyped transcript. Instead, `test` will generate a new transcript that it will pack into `ZipPlusTestTranscript`.
  * This typed transcript is then passed to `evaluate` which returns `ZipPlusProof` along with evaluation result in F itself.
  * `ZipPlusProof` is passed to `verify`.
  * `ZipPlusTestTranscript` can be cloned to do several evaluations relying on the same proof of testing phase.
  * Removed unrelated parameters from `test/evalutate` - e.g. `test` no longer needs a field, and `evaluate` doesn't care about prover hint.
  * Most tests from `open` were moved to `verify` as they require passing through all stages anyway.
  * Removed `read_field_projecting_element` and `MockTranscript` from tests as they are no longer needed.
* Additionally, RAA shuffle seeds made constant as their unpredictibility is not needed for Zip to be secure


maksimryndin (approved):
LGTM

PR #8, opened 2025-10-09
[Chore] Move ConstMontyField to crypto-primitives

frozenspider:
This diff revolves around making `ConstMontyField` a reusable crypto primitive, a part of `crypto-primitives` crate (which will be separated out as some point in future). There are no Zip logic changes.
* Wrapper moved to primitives, implemented missing methods for it, exposed some constants that `crypto_bigint` used to expose.
* Also added some missing pieces to `Int<_>` crypto primitive


maksimryndin (approved):
LGTM

PR #10, opened 2025-10-09
Better crypto primitives

frozenspider:
* Main change - added `Field` wrappers for `ark_ff::fields::Field` and `PrimeField` wrapper for `ark_ff::fields::Fp`, barred behind `ark_ff` feature flag. They can now be used as drop-in replacements for fields with constant modulus.
* Added procedural macro to auto-implement `checked_*` operation that cannot fail - which is a common case when operating on finite fields. 
* `Ring` now requires `Hash`
* Changed `IntRing` - it no longer requires `shl`, `shr` and `rem`, those has been moved to `IntRingWithShifts` and `IntRingWithRem` autoimplemented marker traits.
* Also, `IntRing` is no longer necessarily constant, that's `ConstIntRing` specifically - it also requires `FromStr`. All currently implemented fields and rings are `ConstIntRing`, but the field with dynamic modulus (up next) is not.


maksimryndin (approved):
LGTM

PR #11, opened 2025-10-13
Field with dynamic modulus

frozenspider:
Prime field modulus for evaluation phase is picked by the verifier at the evaluation stage (in practice we use Fiat-Shamir, so it's drafted as a challenge based on transcript).

To make it work, added dynamic field crypto primitive, as well as implementation based on `BoxedMontyForm`.
* Even its zero and one elements cannot be constructed without known modulus, so field API was tweaked to reflect that.
* `Transcribable` and `Transcript` changed to allow reading/writing field with length not known at compile-time. Old behaviour is now a child trait `ConstTranscribable`.
* Ported over Miller-Rabin primality test from original Zinc.
* Added `SimpleSemiring` and `UintSemiring` to make Miller-Rabin generic. These aren't proper crypto primitives.


maksimryndin (approved):
LGTM

PR #12, opened 2025-10-14
Improved RAA code

frozenspider:
Added two new configuration options to RAA code:
* `flip_signs` - if set, signs of code entries on even position would be flipped twice (after repetition, and after first permutation) - as requested by @albert-garreta to make codeword entries smaller on average
* `permute_in_place` - if unset, permutation order would be precomputed and used to copy data in the permuted order. This is faster than permuting in-place using RNG when entries are sufficiently small
* For integers, setting `permute_in_place = false` leads to ~2-2.5x speed boost
* For polynomials, performance gets worse, so `permute_in_place` is still enabled for them


maksimryndin (approved):
LGTM!

PR #13, opened 2025-10-14
Improved Merkle tree

frozenspider:
* Evaluations are now encoded as a table, not as a single vector - this makes working with codewords more natural.
* Removed codewords transposition by using rows representation and getting columns via index.
* Replaced P3's MT implementation with a naive one - turns out it's ~20% faster, especially when using column-first representation
* Removed all P3 supporting code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,3 +30,7 @@ cast_sign_loss = "deny"
 needless_return = "deny"
 panicking_overflow_checks = "deny"
 unwrap_used = "deny"
+
+[profile.release]
+lto = true # Enable Link Time Optimization
+codegen-units = 1 # Slower compilation but potentially better optimization
--- a/crypto-primitives/src/matrix.rs
+++ b/crypto-primitives/src/matrix.rs
@@ -1,4 +1,5 @@
 use alloc::vec::Vec;
+use core::mem::{ManuallyDrop, MaybeUninit};
 
 /// A matrix, rectangular table of values
 pub trait Matrix<T> {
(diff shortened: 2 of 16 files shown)


Cargo.toml
```diff
@@ -30,3 +30,7 @@ cast_sign_loss = "deny"
 needless_return = "deny"
 panicking_overflow_checks = "deny"
 unwrap_used = "deny"
+
+[profile.release]
+lto = true # Enable Link Time Optimization
+codegen-units = 1 # Slower compilation but potentially better optimization
```
cupicmarko, line 36: Just a curiosity, how much does this impact benchmarks? I had to remove this once to improve build times on MacOS, it does sometimes make build slow

crypto-primitives/src/matrix.rs
```diff
@@ -71,40 +76,106 @@ where
...
+
+    pub fn as_rows(&self) -> impl Iterator<Item = &[T]> {
+        self.data.chunks_exact(self.num_cols)
+    }
+
+    pub fn as_rows_mut(&mut self) -> impl Iterator<Item = &mut [T]> {
+        self.data.chunks_exact_mut(self.num_cols)
+    }
+
+    pub fn to_rows_slices(&self) -> Vec<&[T]> {
```
cupicmarko, line 114: Both to_rows_* functions maybe are unnecessary because as_rows* can be used instead? They allocate vectors for slices, slices are no cost, but vectors are allocated each time, and might be dropped after only one iteration

zip-plus/src/utils.rs
cupicmarko, line 152: Not part of this PR, but for a future improvement:
Could we use IntoParallelIterator here, scope.spawn spawns a rayon task per item and it can be very expensive.

cupicmarko (approved):
Seems good, I have left few comments

zip-plus/src/utils.rs
frozenspider: Thanks for the insight! Created #17 for that.

crypto-primitives/src/matrix.rs
frozenspider: They allocate a vector to hold slices, yeah. I've started with `as_rows` but they are often unwieldy to use, especially in tests - it's a recurring pattern that we need to `collect` to be able to address cells by coordinates.
As per [Rust naming guidelines](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv), `as_*` methods are supposed to be free, while`to_*` - expensive

Cargo.toml
frozenspider: Not by a lot, tbh 🙂 
it improves `Zip+/Test: Eval=i32, Cw=i64, Comb=Int<3>, poly_size=2^16` by around ~15%, for other phases impact is negligible

PR #14, opened 2025-10-15
Semiring primitive

frozenspider:
* Added a `Semiring` primitive, essentially a ring with no additive inverse. It fits unsigned integers.
* Added trivial implementations for all unsigned int types.
* Added semiring wrappers `crypto_bigint::Uint` and `bool` types.
* Changed `Int` wrapper to work with `Uint` wrapper instead of with `crypto_bigint::Uint` directly
* Removed `SimpleSemiring` as it's now replaced by full-fledged `Semiring`.
* Added indirection to Miller-Rabin to allow us to implement it for arbitrary semirings.

Zip+ logic itself remains unchanged.

### DIFF
--- a/crypto-primitives/src/semiring/boolean.rs
+++ b/crypto-primitives/src/semiring/boolean.rs
@@ -3,10 +3,13 @@ use core::{
     fmt::{Debug, Display, Formatter, Result as FmtResult},
     hash::{Hash, Hasher},
     iter::{Product, Sum},
-    ops::{Add, AddAssign, Mul, MulAssign, Sub, SubAssign},
+    ops::{Add, AddAssign, Deref, Mul, MulAssign, Sub, SubAssign},
     str::{FromStr, ParseBoolError},
 };
 use num_traits::{CheckedAdd, CheckedMul, CheckedSub, ConstOne, ConstZero, One, Pow, Zero};
+
+#[cfg(feature = "rand")]
+use rand::{distr::StandardUniform, prelude::*};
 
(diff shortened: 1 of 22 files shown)


crypto-primitives/src/semiring/boolean.rs
cupicmarko, line 1: Would [newtype_derive](https://lib.rs/crates/newtype_derive) remove some of the boilerplate code?

crypto-primitives/src/semiring/crypto_bigint_uint.rs
cupicmarko, line 1: I didn't review all the lines here, but I assume it's mostly reshuffled code?

cupicmarko (approved):
LGTM!

crypto-primitives/src/semiring/boolean.rs
frozenspider: Oh nice, let me check

crypto-primitives/src/semiring/crypto_bigint_uint.rs
frozenspider: Yeah, it's just wrappers around `Uint`, code adapted from `Int` wrapper, and some interop with `Int`

crypto-primitives/src/semiring/boolean.rs
frozenspider: Sadly it's hardcoded to use `std::` paths and thus is not compatible with no-std 😞 
Also, this repo hasn't been updated for 9 years.
Otherwise would be nice to remove a bunch of boilerplate from all our types

PR #15, opened 2025-10-16
Boolean evalutaions

frozenspider:
(This is mostly a preparation PR to using binary decomposed polynomials)
* Switched Zip structs and traits to use semirings instead of rings where appropriate.
* RAA code has been split onto RAA and sign-flipping RAA. This is because sign-flipping version require `Neg` trait to be implemented, so it won't work on semirings.
* Zip+ benches changed to use `Boolean` instead of `i8` for evaluation coefficients.
* Additionally, test types for polynomials were changed to match benchmark types of Zip+. (Note that for int test types it's still more convenient to use types that differ to benchmarks)

### DIFF
--- a/zip-plus/benches/zip_benches.rs
+++ b/zip-plus/benches/zip_benches.rs
@@ -35,7 +35,6 @@ type Code = RaaCode<BenchZipTypes, 4>;
 const RAA_CONFIG: RaaConfig = RaaConfig {
     check_for_overflows: false,
     permute_in_place: false,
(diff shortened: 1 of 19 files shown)


cupicmarko (approved):
Great!

Food for thought:
Would splitting Ring, Semiring and other types into:
- Abelian group
- Monoid
- Ring

make sense?

frozenspider:
@cupicmarko thanks for the idea, created #18 to discuss this further

PR #16, opened 2025-10-19
Refactor polynomials

frozenspider:
Removed `EvalR` and `CwR` types from `ZipTypes`. Instead, `ZipTypes` is now parameterized by polynomial degree. Added separate trait `ConstCoeffBitWidth` for polynomials that need to know their coefficient bit width (`Eval` and `Cw`)
Also replaced trait `Polynomial` with `EvaluatablePolynomial` with different type signature and intentions. It can be implemented multiple times if needed.

### DIFF
--- a/zip-plus/src/poly/dense.rs
+++ b/zip-plus/src/poly/dense.rs
@@ -1,4 +1,4 @@
-use super::{EvaluationError, Polynomial};
+use super::{EvaluatablePolynomial, EvaluationError, Polynomial};
 use crate::{
     pcs::structs::{MulByScalar, ProjectableToField},
     traits::{ConstTranscribable, FromRef, Named},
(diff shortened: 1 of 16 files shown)


zip-plus/src/poly/dense.rs
```diff
@@ -371,6 +356,39 @@ where
...
+        if point.len() != DEGREE {
+            return Err(EvaluationError::WrongPointWidth {
+                expected: DEGREE,
+                actual: point.len(),
+            });
+        }
+
+        // A trivial implementation of a polynomial evaluation at a given point.
+        let mut result = self.coeff_0.clone();
+        for (coeff, scalar) in self.coeffs.iter().zip(point.iter()) {
```
cupicmarko, line 376: Something like this can still be safe and remove hidden checks with zip:
```rust
let mut result = self.coeff_0.clone();
unsafe {
    for i in 0..DEGREE {
        let coeff = self.coeffs.get_unchecked(i);
        let s = point.get_unchecked(i);
        let t = coeff.mul_by_scalar(s).ok_or(EvaluationError::Overflow)?;
        result = acc.checked_add(&t).ok_or(EvaluationError::Overflow)?;
    }
}
```

cupicmarko (approved):
Looks really good!

zip-plus/src/poly/dense.rs
frozenspider: Tried that, but to no performance benefit.
Turns out Rust compiler is smart enough to do it on its own 🙂 
https://godbolt.org/z/G1nTh5h4b

PR #19, opened 2025-10-24
Better parallelism

### LINKED ISSUE #17: Replace rayon::scope with into_par_iter
https://github.com/NethermindEth/zinc-plus/pull/13#discussion_r2446218437
> Not part of this PR, but for a future improvement:
> Could we use IntoParallelIterator here, scope.spawn spawns a rayon task per item and it can be very expensive.

frozenspider:
Turns out that thread limiting approach used in `parallelize` adds more overhead than it removes.
Replaced our `parallelize` and `parallelize_*` methods with `cfg_*` macros from `ark-std`.
As a result, code was simplified and `encode` performace for primitive types improved by ~40% with `parallel` feature enabled.

Fixes #17

### DIFF
--- a/zip-plus/src/merkle.rs
+++ b/zip-plus/src/merkle.rs
@@ -1,5 +1,5 @@
 use crate::{add, traits::ConstTranscribable};
-use ark_std::cfg_into_iter;
+use ark_std::cfg_iter_mut;
 use itertools::Itertools;
 use std::{
     fmt,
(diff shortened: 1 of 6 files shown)


cupicmarko (approved):
LGTM!

I expected this to show great improvement, because "the old way" was mostly reusing sequential iterator and zipping them, I don't honestly know how much can a compiler can optimize it. This a way to go even if performance improved only slightly because it's more readable and idiomatic

PR #20, opened 2025-10-27
BLAKE3 natural Merkle tree implementation

frozenspider:
BLAKE3 hash for arbitrary input works by splitting initial input into 1024-byte chunks, hashing each of them and merging those hashes together, thus building a binary tree on top of initial chunks.

This structure in essence is a Merkle tree on its own, and `blake3` crate provides access to low-level intrinsics of this process through [hazmat](https://docs.rs/blake3/latest/blake3/hazmat/index.html) submodule.

To avoid unnecessary overhead when building MT for Zinc, we abuse the tree structure of BLAKE3 to construct a tree on top of hashed leaves. However, MT building time is heavily dominated by leaves hashing, so this change leads to a meager ~5% performance increase of MT construction.

As a result, when committing polynomial of size 2^16 and `i64` codeword entries, 98% of overall MT building time is leaves hashing, and 92% of that is spent doing `blake3::Hasher::update`. Percents are even higher when hashing polynomials.
Given that we're also making good use of `rayon`, current solutions is close to optimal under the assumption that we keep BLAKE3 as a hash function for MT construction.


zip-plus/src/merkle.rs
```diff
@@ -79,217 +74,251 @@ impl MerkleTree {
...
+            siblings.push(current_layer[sibling_index].clone());
         }
+
+        current_index /= 2;
+        layer_idx += 1;
+        current_layer = &layers[layer_idx];
     }
-    MerkleTreeInner { layers }
+
+    // Reverse to get bottom-up order (leaf to root) as expected by verification
```
cupicmarko: Is this comment correct? The path is aldready bottom-up if I understand code correctly

cupicmarko (approved):
LGTM!

zip-plus/src/merkle.rs
frozenspider: Right, forgot to remove this comment after logic change, thanks!

PR #21, opened 2025-10-30
Remove coeff_0 in DensePolynomial and DEGREE from ZipTypes

frozenspider:
* Due to inability to limitation of stable rust (cannot use math operations on generic constants as array size), `DensePolynomial` used `coeff_0` and `coeffs: [_; DEGREE]`. However, using `coeffs: [_; DEGREE_PLUS_ONE]` is much more simple and convenient workaround.

* `ZipTypes`,`LinearCode` and all Zip phases were parameterized by `DEGREE`, but that was largely a leftover from a previous design iteration. The only place it was actually used was to pass it to `validate_input`. Instead, re-introduced a simplified version of `Polynomial` trait that's aware of its own degree. (Luckily, math operations on generic constants are allowed there.)

This refactoring does not affect behaviour, benchmarked performance is also largely unchanged.

### DIFF
--- a/zip-plus/benches/zip_benches.rs
+++ b/zip-plus/benches/zip_benches.rs
@@ -17,7 +17,7 @@ use zip_plus::{
 const INT_LIMBS: usize = U64::LIMBS;
 
 struct BenchZipTypes {}
-impl ZipTypes<0> for BenchZipTypes {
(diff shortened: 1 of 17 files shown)


cupicmarko (approved):
LGTM!

PR #23, opened 2025-11-04
An attempt to simplify the field stuff

ElijahVlasov:
1. Remove the unused field implementations. In particular the arkworks' fields and constant field.
2. Make the main field implementation have limb size known in compile time. (`BoxedMontyField` -> `MontyField`). That doesn't worsen the benches nor does it break any tests.
3. Remove the `Rem`-related traits from the field trait stack as it's a pretty alien operation for fields.


frozenspider:
Hey @ElijahVlasov.
The initial idea for `crypto-primitives` crate was to split it off into a separate library, which is being done here: https://github.com/NethermindEth/crypto-primitives/pull/1
After it merges, you can declare a dependency on it instead of removing unused code. I agree it makes sense to carry over `MontyForm` in addition to `BoxedMontyForm` though.

ElijahVlasov:
> Hey @ElijahVlasov. The initial idea for `crypto-primitives` crate was to split it off into a separate library, which is being done here: [NethermindEth/crypto-primitives#1](https://github.com/NethermindEth/crypto-primitives/pull/1) After it merges, you can declare a dependency on it instead of removing unused code. I agree it makes sense to carry over `MontyForm` in addition to `BoxedMontyForm` though.

Does it really make sense? It seems to me that `crypto-primitives` is specific for zinc+ and isn't super generic

PR #24, opened 2025-11-10
Migrate to crypto-primitives repo dependency

ElijahVlasov:
`crypto-primitives*` crates have been moved to https://github.com/NethermindEth/crypto-primitives.


cupicmarko (approved):
Looks good, I tried to cross-reference what was moved into crypto-primitives and everything looks good, I only suggest adding URL to crypto-primitive repository to merge commit message to make git log clearer.

PR #25, opened 2025-11-10
Extract `poly` and `transcript` into separate crates

ElijahVlasov:
Separate polynomial and transcript stuff from the `zip-plus` crate.

The sum check protocol depends on multilinear polynomials and transcript while being independent of the PCS, so it's reasonable to have them in appropriate creates.


cupicmarko (approved):
Looks good! Just a curiosity, are other modules going to be split off into crates?

ElijahVlasov:
> Looks good! Just a curiosity, are other modules going to be split off into crates?

Is anything worth splitting off? It seems that the rest of zip-plus belongs to zip-plus just fine.

PR #26, opened 2025-11-12
Sumcheck on a field over a (possibly) random prime

ElijahVlasov:
Sumcheck over a (possibly) random field.

Ported the implementation from https://github.com/NethermindEth/zinc/tree/main/src/sumcheck.


piop/src/sumcheck/verifier.rs
```diff
@@ -0,0 +1,330 @@
...
+            let evaluations = &verifier_state.polynomials_received[i];
+            if evaluations.len() != verifier_state.max_multiplicands + 1 {
+                return Err(SumCheckError::MaxDegreeExceeded);
+            }
+
+            let p0 = &evaluations[0];
+            if verifier_state.max_multiplicands > 0 {
+                let p1 = &evaluations[1];
+                if p0.clone() + p1.clone() != expected {
+                    return Err(SumCheckError::SumCheckFailed(
```
osdnk, line 129: There is the trick (cf. 3.1 of https://eprint.iacr.org/2024/108.pdf) that you don't need to send `0` (or some) evaluation of the polynomial, but you may recompute it from asserting that the checks go through.

piop/src/sumcheck/verifier.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+}
+
+/// interpolate the *unique* univariate polynomial of degree *at most*
+/// p_i.len()-1 passing through the y-values in p_i at x = 0,..., p_i.len()-1
+/// and evaluate this  polynomial at `eval_at`. In other words, efficiently
+/// compute  \sum_{i=0}^{len p_i - 1} p_i[i] * (\prod_{j!=i} (eval_at -
+/// j)/(i-j))
+#[allow(clippy::cast_possible_wrap)]
+pub(crate) fn interpolate_uni_poly<F: FromPrimitiveWithConfig>(
```
osdnk, line 162: Wouldn't it be more efficient to send just the coefficients of the polynomial and then evaluate at the point of the Boolean hypercube on the verifier side? Assuming we are interested only in the boolean hypercube evaluating at 0 is easy (take ct), so is at 1 (sum coeffs). The complexity of the current impl is quadratic in the number of coefficients. This round-trip to the eval domain and back might be unnecessary.

ElijahVlasov: We're interested in evaluating on a random challenge as well. The values on 0 and 1 are coming from the prover side as we send evaluation p(0), p(1), ..., p(deg p -1) now

piop/src/sumcheck/verifier.rs
ElijahVlasov: Yes, we have a task in the control centre for applying optimisations like in that paper. I think it's out of the scope of this particular PR. 

Sumcheck benches have to be set up before applying any optimisations as well.

piop/src/sumcheck/verifier.rs
osdnk: I don't get the point. I think it's just unnecessarily complicated

ElijahVlasov: In latticefold we had a coefficient form based sum-check which was far worse than this implementation. Not necessarily is it better in zinc context but we assume it is :)

transcript/src/traits.rs
```diff
@@ -76,13 +76,44 @@ pub trait Transcript {
     /// current transcript state, updating it.
     fn get_challenge<T: ConstTranscribable>(&mut self) -> T;
 
+    fn get_field_challenge<F: PrimeField>(&mut self, cfg: &F::Config) -> F
+    where
+        F::Inner: ConstTranscribable,
+    {
+        let random_inner = self.get_challenge();
+
+        F::new_unchecked_with_cfg(random_inner, cfg)
```
frozenspider: new_unchecked_with_cfg expects representation to be a valid field element, but a random Inner doesn't have to be a valid field element in a general case (e.g. say we declare M31 field that's represented as u32).
Maybe make it a loop until challenge < modulus?

piop/src/sumcheck/tests/utils.rs
```diff
@@ -0,0 +1,110 @@
...
+use crypto_primitives::PrimeField;
+use rand::{Rng, RngCore};
+use std::sync::Arc;
+use zinc_poly::{mle::DenseMultilinearExtension, utils::ArithErrors};
+
+#[allow(clippy::arithmetic_side_effects)]
+#[allow(clippy::type_complexity)]
+pub(crate) fn rand_poly<F: PrimeField + Random, Rn: RngCore>(
+    nv: usize,
+    num_multiplicands_range: (usize, usize),
```
frozenspider: This is used only as a half-open range, so just take it as such?
```suggestion
    num_multiplicands_range: Range<usize>,
```

piop/src/sumcheck/tests/utils.rs
```diff
@@ -0,0 +1,110 @@
...
+    ArithErrors,
+> {
+    let mut sum = F::zero_with_cfg(config);
+    let mut mles = vec![];
+    let mut products = Vec::with_capacity(num_products);
+    let mut degree = 0;
+    let mut current_mle_index = 0;
+    for _ in 0..num_products {
+        let num_multiplicands =
+            rng.random_range(num_multiplicands_range.0..num_multiplicands_range.1);
```
frozenspider: With the change above this would become:
```suggestion
            rng.random_range(num_multiplicands_range.clone());
```

piop/src/sumcheck/tests/utils.rs
```diff
@@ -0,0 +1,110 @@
...
+            e.push(val.clone());
+            product *= &val;
+        }
+        sum += &product;
+    }
+
+    let list = multiplicands
+        .into_iter()
+        .map(|x| {
+            Arc::new(DenseMultilinearExtension::from_evaluations_vec(
```
frozenspider: There doesn't seem to be a reason to have `Arc` here, all it does is getting uwrapped immediately in `rand_poly`

piop/src/sumcheck/tests/utils.rs
```diff
@@ -0,0 +1,110 @@
...
+    vals: &[F],
+    products: &[(F, Vec<usize>)],
+    config: F::Config,
+) -> F {
+    let mut result = F::zero_with_cfg(&config);
+    for (coef, indices) in products {
+        let mut term = coef.clone();
+        for &i in indices {
+            term *= &vals[i];
+        }
```
frozenspider: (Nit)
```suggestion
        let term = coef.clone() * indices.iter().map(|&i| &vals[i]).product::<F>();
```

piop/src/sumcheck/tests/utils.rs
```diff
@@ -0,0 +1,110 @@
...
+    nv: usize,
+    degree: usize,
+    rng: &mut Rn,
+    config: F::Config,
+) -> (Vec<Arc<DenseMultilinearExtension<F>>>, F) {
+    let start = start_timer!(|| "sample random mle list");
+    let mut multiplicands = Vec::with_capacity(degree);
+    for _ in 0..degree {
+        multiplicands.push(Vec::with_capacity(1 << nv));
+    }
```
frozenspider: In general it's more preferrable to use map/collect one-liners, but in this case there's a convenient macro:
```suggestion
    let mut multiplicands = vec![Vec::with_capacity(1 << nv); degree];
```

piop/src/sumcheck/prover.rs
```diff
@@ -0,0 +1,176 @@
...
+    pub num_vars: usize,
+    /// Max degree
+    pub max_degree: usize,
+    /// The current round number
+    pub round: usize,
+}
+
+impl<F: PrimeField> IPForMLSumcheck<F>
+where
+    for<'a> F: MulByScalar<&'a F>,
```
frozenspider: Nit: Where there are multiple bounds and `where` clause is used, it's (arguably) better to move all bounds into the `where` clause rather than splitting some of them to be inlined. Feel free to ignore this though.

piop/src/sumcheck/prover.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        #[cfg(not(feature = "parallel"))]
+        let zeros = scratch();
+        #[cfg(feature = "parallel")]
+        let zeros = scratch;
+
+        let summer = cfg_into_iter!(0..1 << (nv - i)).fold(zeros, |mut s, b| {
+            let index = b << 1;
+
+            s.vals0
```
frozenspider, line 130: Once you have benches set, could please try getting rid of `vals0` and `vals1` fields in the structs, replacing them with
```rust
let vals0: Vec<_> = polys.iter().map(|poly| poly[index].clone()).collect();
let vals1: Vec<_> = polys.iter().map(|poly| poly[index + 1].clone()).collect();
```
? My bet is that it won't affect running time, but better safe than sorry

piop/src/sumcheck/prover.rs
```diff
@@ -0,0 +1,176 @@
...
+                .zip(s.levals.iter())
+                .for_each(|(e, l)| *e += l);
+            s
+        });
+
+        // Rayon's fold outputs an iter which still needs to be summed over
+        #[cfg(feature = "parallel")]
+        let evaluations = summer.map(|s| s.evals).reduce(
+            || vec![zero.clone(); degree + 1],
+            |mut evaluations, levals| {
```
frozenspider: A lie! 🙂 This is `evals`, not `levals`

piop/src/sumcheck/prover.rs
```diff
@@ -0,0 +1,176 @@
...
+                    }
+                    *eval_point = comb_fn(&s.vals);
+                }
+            }
+
+            s.evals
+                .iter_mut()
+                .zip(s.levals.iter())
+                .for_each(|(e, l)| *e += l);
+            s
```
frozenspider, line 166: It seems that the only thing we pass around meaningfully is `evals`, so this loop could be reworked to map/reduce - maybe even without `#[cfg(feature = "parallel")]`. Would help to get benchmarks up an running first though.

piop/src/sumcheck/tests.rs
```diff
@@ -0,0 +1,637 @@
...
+};
+
+use super::{MLSumcheck, SumcheckProof};
+
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+struct Params {}
+
+const N: usize = 2;
+
+impl ConstMontyParams<2> for Params {
```
frozenspider: Nit: Use `const_monty_params!` macro?

piop/src/sumcheck/verifier.rs
```diff
@@ -0,0 +1,330 @@
...
+    //  - 2^122 < factorial(33) < 2^123
+    // so we will be able to compute the ratio
+    //  - for len <= 20 with i64
+    //  - for len <= 33 with i128
+    //  - for len >  33 with BigInt
+    if p_i.len() <= 20 {
+        let last_denom: F = F::from_with_cfg(u64_factorial(len - 1), config);
+
+        let mut ratio_numerator = 1i64;
+        let mut ratio_enumerator = 1u64;
```
frozenspider: Nit: Could we pick better terms than `numerator` and `enumerator`? Their combination is very confusing 🫤

piop/src/sumcheck/verifier.rs
```diff
@@ -0,0 +1,330 @@
...
+                / (last_denom.clone() * ratio_numerator_f * &evals[i]);
+            res += &(p_i[i].clone() * x);
+
+            // compute ratio for the next step which is current_ratio * -(len-i)/i
+            if i != 0 {
+                ratio_numerator *= -(len as i128 - i as i128);
+                ratio_enumerator *= i as u128;
+            }
+        }
+    } else {
```
frozenspider, line 269: Once we have benches, it's worth checking if we're even winning anything with specialized branches above

piop/src/sumcheck/verifier.rs
```diff
@@ -0,0 +1,330 @@
...
+
+    res
+}
+
+/// compute the factorial(a) = 1 * 2 * ... * a
+#[inline]
+fn field_factorial<F: FromWithConfig<u64>>(a: usize, config: &F::Config) -> F {
+    let mut res = F::one_with_cfg(config);
+    for i in 1..=(a as u64) {
+        res *= F::from_with_cfg(i, config);
```
frozenspider: Nit: Use `.map().product()`
(same for u128/u64)

piop/src/sumcheck/verifier.rs
```diff
@@ -0,0 +1,330 @@
...
+    let mut res = 1u128;
+    for i in 1..=a {
+        res *= i as u128;
+    }
+    res
+}
+
+/// compute the factorial(a) = 1 * 2 * ... * a
+#[inline]
+fn u64_factorial(a: usize) -> u64 {
```
frozenspider: Should be made generic over any semiring I think

piop/src/sumcheck/verifier.rs
```diff
@@ -0,0 +1,330 @@
+#![allow(clippy::arithmetic_side_effects)]
```
frozenspider: Bad practice! Please annotate functions separately, preferably with a comment why is it safe in that particular case

piop/src/sumcheck/verifier.rs
```diff
@@ -0,0 +1,330 @@
...
+            expected_evaluation: expected,
+        })
+    }
+
+    /// simulate a verifier message without doing verification
+    ///
+    /// Given the same calling context, `transcript_round` output exactly the
+    /// same message as `verify_round`
+    #[inline]
+    pub fn sample_round(transcript: &mut impl Transcript, config: &F::Config) -> VerifierMsg<F>
```
frozenspider: Any reason we might want this, as opposed to using `transcript.get_field_challenge(config)` directly?

piop/src/sumcheck/verifier.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    /// Run verifier at current round, given prover message
+    ///
+    /// Normally, this function should perform actual verification. Instead,
+    /// `verify_round` only samples and stores randomness and perform
+    /// verifications altogether in `check_and_generate_subclaim` at
+    /// the last step.
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn verify_round(
```
frozenspider: Maybe make all these methods on `VerifierState`?

piop/src/sumcheck.rs
```diff
@@ -0,0 +1,169 @@
...
+        comb_fn: impl Fn(&[F]) -> F + Send + Sync,
+        config: F::Config,
+    ) -> (SumcheckProof<F>, ProverState<F>)
+    where
+        F::Inner: ConstTranscribable,
+        for<'a> F: MulByScalar<&'a F>,
+    {
+        let mut buf = vec![0; F::Inner::NUM_BYTES];
+        F::from_with_cfg(nvars as u64, &config);
+        F::from_with_cfg(degree as u64, &config);
```
frozenspider: These two lines are noop

piop/src/sumcheck.rs
```diff
@@ -0,0 +1,169 @@
...
+        transcript.absorb_random_field(&degree_field, &mut buf);
+        if nvars == 0 {
+            let base_vals: Vec<F> = mles
+                .iter()
+                .map(|m| {
+                    m.evaluate(&[], F::zero_with_cfg(&config))
+                        .expect("eval at empty point")
+                })
+                .collect();
+            let _sum_at_empty = comb_fn(&base_vals);
```
frozenspider: I wonder why is this needed, do we rely on checks done inside or something? Otherwise it looks like a noop

piop/src/sumcheck/prover.rs
```diff
@@ -0,0 +1,176 @@
...
+            round: 0,
+        }
+    }
+
+    /// receive message from verifier, generate prover message, and proceed to
+    /// next round
+    ///
+    /// Adapted Jolt's sumcheck implementation
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn prove_round(
```
frozenspider, line 55: Maybe make this a method on `ProverState`, together with `prover_init` (changing it to just `init`)?

transcript/src/field/const_monty.rs
```diff
@@ -9,7 +9,7 @@ impl<Mod: ConstMontyParams<LIMBS>, const LIMBS: usize> ConstTranscribable
     const NUM_BYTES: usize = Uint::<LIMBS>::NUM_BYTES;
 
     fn read_transcription_bytes(bytes: &[u8]) -> Self {
-        ConstMontyForm::from_montgomery(Uint::read_transcription_bytes(bytes).into_inner())
+        ConstMontyForm::new(&Uint::read_transcription_bytes(bytes).into_inner())
```
frozenspider: Shouldn't a transcription entry already be in Montgomery form?

frozenspider (changes requested):

transcript/src/traits.rs
ElijahVlasov: Yeah, in the crypto-primitives it's not super consistent.

For `ConstMontyField` we have
` type Inner = ConstMontyForm<Mod, LIMBS>;`

Whereas for `BoxedMontyField` it's
` type Inner = BoxedUint;`

Can we enforce that for all field types `Inner` is the monty form from the `crypto_bigint`?

ElijahVlasov: > Maybe make it a loop until challenge < modulus

Not a big fan of that. I'd rather take a random Uint and mod it. That'd get us a consistent running time of `get_field_challenge`. 

Tell me what you think.

piop/src/sumcheck/tests/utils.rs
ElijahVlasov: Using vec! macro here isn't great as it clones under the hood and `Vec::clone` doesn't respect capacity.
https://stackoverflow.com/questions/77136991/why-does-vecvecwith-capacityn-create-0-capacity-for-the-sub-vectors

transcript/src/traits.rs
frozenspider: The point of `new_unchecked` is precisely that it doesn't check if the value is valid, this is by design. Sometimes the check is unavoidable - as it is in `ConstMontyField` - but that's not what API suggests.

> Not a big fan of that. I'd rather take a random Uint and mod it. That'd get us a consistent running time of get_field_challenge.

Looks like this is a trade-off between running time being consistent and distribution being uniform, if latter doesn't matter, doing mod is fine.

piop/src/sumcheck/tests/utils.rs
frozenspider: Interesting. In this case I would go with `(0..degree).map(|_| Vec::with_capacity(1 << nv))).collect_vec()`

ElijahVlasov: Not that it matters much as it's in tests. If it's ever to be copied into non-test parts of the code let's simplify the job for whoever does that :)

piop/src/sumcheck/verifier.rs
ElijahVlasov: There's no much reason behind it. The sumcheck code we have gets moved across codebases quite a lot and the get challenge interface differs from project to project. I like the idea of having proof system specific logic encapsulated in designated methods while keeping the rest of the code copy-pasteable as much as possible. If you insist I'll inline :)

frozenspider: I don't have a strong opinion on this, just looks unnecessary to me.
Leaving it up to you.

ElijahVlasov: Would it make sense to strip VerifierMsg off too then?

ElijahVlasov: If we're to remove sample_round

frozenspider: I think so, we're just immediately stripping `randomness` out of it, and it's not like comments/naming on `VerifierMsg` adds a lot of context to what's going on

piop/src/sumcheck.rs
ElijahVlasov: Well, a (sane) person shouldn't attempt to sumcheck constants anyway. Nuked :)

ElijahVlasov:
Depends on https://github.com/NethermindEth/crypto-primitives/pull/10

transcript/src/traits.rs
ElijahVlasov: I ended up going for the generate a random uint and then reduce approach

ElijahVlasov: Which was made possible by https://github.com/NethermindEth/crypto-primitives/pull/10

transcript/src/field/const_monty.rs
ElijahVlasov: This can be nuked I guess

piop/src/sumcheck/verifier.rs
```diff
@@ -0,0 +1,301 @@
...
+
+/// Interpolate the *unique* univariate polynomial of degree *at most*
+/// p_i.len()-1 passing through the y-values in p_i at x = 0,..., p_i.len()-1
+/// and evaluate this  polynomial at `eval_at`. In other words, efficiently
+/// compute  $\sum_{i=0}^{len\ p_i - 1} p_i\[i\] * (\prod_{j!=i}
+/// (\text{eval\\_at} - j)/(i-j))$.
+// All the arithmetic ops in the function
+// are made sure to not overflow.
+#[allow(clippy::arithmetic_side_effects)]
+#[allow(clippy::cast_possible_wrap)]
```
frozenspider: Nit: Could be specified in one instruction:
```rust
#[allow(clippy::arithmetic_side_effects, clippy::cast_possible_wrap)]
```

Cargo.toml
```diff
@@ -11,8 +11,7 @@ license = "Apache-2.0"
 publish = false
 
 [workspace.dependencies]
-crypto-primitives = { git = "https://github.com/NethermindEth/crypto-primitives.git", branch = "main", features = ["std", "ark_std", "crypto_bigint", "rand"]  }
-crypto-primitives-proc-macros = { git = "https://github.com/NethermindEth/crypto-primitives.git", branch = "main" }
+crypto-primitives = { git = "https://github.com/NethermindEth/crypto-primitives.git", branch = "ElijahVlasov/field-constructor", features = ["std", "ark_std", "crypto_bigint", "rand"]  }
```
frozenspider: Just a reminder to merge https://github.com/NethermindEth/crypto-primitives/pull/10 first and change this dependency before merging this PR

frozenspider (approved):

PR #27, opened 2025-11-17
Introduce traits `InnerProduct` and `InnerProductWithConfig`

ElijahVlasov:
There's a loose abstraction in both the paper and the implementation of evaluating a univariate polynomial on a vector of points instead of denoting it as an inner product. This PR fixes that in the code and unifies various kinds of inner products by introducing the trait `InnerProduct`.

Along the way, the dense univariate polys are moved into `univariate::dense` module and the interface of the univariate evaluation is slightly changed. Now, it doesn't expect a slice for evaluation which typically was a slice of powers but rather a single point whose powers it computes under the hood.

### DIFF
--- a/poly/src/lib.rs
+++ b/poly/src/lib.rs
@@ -12,13 +12,13 @@ pub trait Polynomial<C> {
 }
 
 pub trait EvaluatablePolynomial<C, S, Out>: Polynomial<C> {
-    /// Evaluates the polynomial at the given point, treating point `[p_0, p_1,
-    /// p_2, ...]` as `[x, x^2, x^3, ..., x^DEGREE_BOUND]`, thus it returns
-    /// `a_0 + (a_1 * p_0) + (a_2 * p_1) + ... + (a_DEGREE_BOUND *
-    /// p_{DEGREE_BOUND - 1})`.
-    // Note: we can't reference Self::DEGREE_BOUND here, type parameters may not be
-    // used in const expressions. As such, no point in declaring Self: Polynomial.
-    fn evaluate_at_point(&self, point: &[S]) -> Result<Out, EvaluationError>;
+    /// The type of points a polynomial can be evaluated on.
+    /// For univariate polynomials this typically is `C`,
+    /// for multivariate this is `[C]`.
+    type EvaluationPoint: ?Sized;
+
+    /// Evaluates the polynomial at the given point.
+    fn evaluate_at_point(&self, point: &Self::EvaluationPoint) -> Result<Out, EvaluationError>;
 }
 
 pub trait ConstCoeffBitWidth {
@@ -31,4 +31,6 @@ pub enum EvaluationError {
     WrongPointWidth { expected: usize, actual: usize },
     #[error("Evaluation failed due to overflow")]
     Overflow,
+    #[error("Empty polynomials are not allowed to be evaluate")]
+    EmptyPolynomial,
 }
--- a/poly/src/lib.rs
+++ b/poly/src/lib.rs
@@ -1,5 +1,5 @@
-pub mod dense;
 pub mod mle;
+pub mod univariate;
 pub mod utils;
 pub mod zero_degree;
 
--- a/poly/src/univariate/dense.rs
+++ b/poly/src/univariate/dense.rs
@@ -1,5 +1,4 @@
-use super::{ConstCoeffBitWidth, EvaluatablePolynomial, EvaluationError};
-use crate::Polynomial;
+use crate::{ConstCoeffBitWidth, EvaluatablePolynomial, EvaluationError, Polynomial};
 use crypto_primitives::{FromWithConfig, IntoWithConfig, PrimeField, Ring, Semiring};
 use itertools::Itertools;
 use num_traits::{CheckedAdd, CheckedMul, CheckedNeg, CheckedSub, One, Zero};
@@ -16,7 +15,6 @@ use zinc_utils::{
     from_ref::FromRef, mul_by_scalar::MulByScalar, named::Named,
     projectable_to_field::ProjectableToField, sub,
 };
-// TODO: rename to univariate?
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct DensePolynomial<R, const DEGREE_PLUS_ONE: usize> {
--- a/poly/src/univariate/dense.rs
+++ b/poly/src/univariate/dense.rs
@@ -1,4 +1,5 @@
 use crate::{ConstCoeffBitWidth, EvaluatablePolynomial, EvaluationError, Polynomial};
+use core::slice;
 use crypto_primitives::{FromWithConfig, IntoWithConfig, PrimeField, Ring, Semiring};
 use itertools::Itertools;
 use num_traits::{CheckedAdd, CheckedMul, CheckedNeg, CheckedSub, One, Zero};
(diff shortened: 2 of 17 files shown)


utils/src/inner_product.rs
```diff
@@ -0,0 +1,143 @@
...
+{
+    type Output = Lhs;
+    type Config = Lhs::Config;
+
+    fn inner_product(&self, rhs: &[Rhs], config: &Lhs::Config) -> Self::Output {
+        self.as_slice().inner_product(rhs, config)
+    }
+}
+
+// TODO: can these two merged into one?
```
frozenspider: You can do it the dirty way, with some inlined `#[cfg(feature = "parallel")]` and the opposite, similarly to how prover does it.
Well, at least iter can be completely abstracted away with `cfg_iter!`...

Something like:
```rust
let half_baked = cfg_iter!(lhs)
    .zip(rhs)
    .map(|(lhs, rhs)| {
        lhs.mul_by_scalar(rhs)
            .expect("Cannot multiply an element by a coefficient")
    });

#[cfg(not(feature = "parallel"))]
half_baked
    .reduce(|acc, product| add!(acc, &product))
    .unwrap_or(zero())

#[cfg(feature = "parallel")]
half_baked
    .reduce(zero, |acc, product| add!(acc, &product))
```

It might be that we actually don't want to parallelize this though, as the overhead might outweigh the benefits. Have you tried benchmarking both approaches?

utils/src/inner_product.rs
```diff
@@ -0,0 +1,143 @@
...
+
+impl<Lhs, Rhs> InnerProductWithConfig<Rhs> for &[Lhs]
+where
+    Lhs: PrimeField + CheckedAdd + for<'z> MulByScalar<&'z Rhs>,
+    Rhs: Send + Sync,
+{
+    type Output = Lhs;
+    type Config = Lhs::Config;
+
+    fn inner_product(&self, rhs: &[Rhs], config: &Lhs::Config) -> Self::Output {
```
frozenspider: We don't need a full config (nor do we need to know its type), all we need is a cloneable zero element.
If we don't mind additional parameter, we can unite both `InnerProduct`s - this would avoid a necessity to call `InnerProductWithConfig::inner_product` explicitly.

utils/src/inner_product.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+// TODO: can these two merged into one?
+
+#[cfg(not(feature = "parallel"))]
+fn inner_product_aux<Lhs, Rhs, Z>(lhs: &[Lhs], rhs: &[Rhs], zero: Z) -> Lhs
+where
+    Lhs: Clone + CheckedAdd + for<'z> MulByScalar<&'z Rhs>,
+    Z: Fn() -> Lhs,
```
frozenspider: Making this a function doesn't achieve much, I'd rather just take a zero reference and clone is as needed

utils/src/inner_product.rs
```diff
@@ -0,0 +1,143 @@
...
+{
+    lhs.iter()
```
frozenspider: I think we might want to add checks for matching length, and (maybe?) return `Result` rather than panicking

poly/src/lib.rs
```diff
@@ -12,13 +12,13 @@ pub trait Polynomial<C> {
...
-    // Note: we can't reference Self::DEGREE_BOUND here, type parameters may not be
-    // used in const expressions. As such, no point in declaring Self: Polynomial.
-    fn evaluate_at_point(&self, point: &[S]) -> Result<Out, EvaluationError>;
+    /// The type of points a polynomial can be evaluated on.
+    /// For univariate polynomials this typically is `C`,
+    /// for multivariate this is `[C]`.
+    type EvaluationPoint: ?Sized;
+
+    /// Evaluates the polynomial at the given point.
+    fn evaluate_at_point(&self, point: &Self::EvaluationPoint) -> Result<Out, EvaluationError>;
```
frozenspider, line 21: It seems that after the introduction of `InnerProduct`, the only place where `evaluate_at_point` is still being called/used is in `DensePolynomial::prepare_projection`, and it's not used for other types at all.
If we expect it to be the case going forward, we should probably remove this trait and make `evaluate_at_point` a private method on `DensePolynomial`.

frozenspider (changes requested):
@ElijahVlasov, just to confirm, did you run benchmarks (both parallel and sequential) to make sure performance didn't degrade as the result of this PR?
We should be doing this on each one that changes existing code

poly/src/lib.rs
ElijahVlasov: I wanted to turn the interpolate thingie from the sumcheck pr into an instance of `EvaluateablePolynomial`

utils/src/inner_product.rs
ElijahVlasov: So. I tried benchmarking on `--features parallel` the current `HEAD` (no parallelisation in `inner_product`) against the `HEAD~1` (parallelisation in `inner_product`) and the parallel inner product falls behind. So I'm removing that

ElijahVlasov:
> @ElijahVlasov, just to confirm, did you run benchmarks (both parallel and sequential) to make sure performance didn't degrade as the result of this PR? We should be doing this on each one that changes existing code

Confirmed :)

zip-plus/src/pcs/phase_test.rs
```diff
@@ -37,15 +37,12 @@ impl<Zt: ZipTypes, Lc: LinearCode<Zt>> ZipPlus<Zt, Lc> {
...
+            let evals: Vec<_> = poly
                 .evaluations
                 .iter()
                 .map(Zt::Comb::from_ref)
-                .map(|p| {
-                    p.evaluate_at_point(&alphas)
-                        .expect("Failed to evaluate polynomial")
-                })
-                .collect_vec();
+                .map(|p| p.inner_product(&alphas, Zt::CombR::from(0)))
```
frozenspider, line 44: (Here and in other places)
`Zt::CombR::ZERO`, or `ConstZero::ZERO`

utils/src/inner_product.rs
```diff
@@ -0,0 +1,100 @@
...
+{
+    type Output = Lhs;
+
+    fn inner_product(&self, rhs: &[Rhs], zero: Lhs) -> Result<Self::Output, InnerProductError> {
+        let half_baked = self
+            .iter()
+            .zip(rhs)
+            .map(|(lhs, rhs)| lhs.mul_by_scalar(rhs).ok_or(InnerProductError::Overflow));
+
+        half_baked
```
frozenspider: You don't need an intermediate `half_baked` anymore

poly/src/univariate/dense.rs
```diff
@@ -365,24 +367,23 @@ impl<R: Semiring, C, const DEGREE_PLUS_ONE: usize> EvaluatablePolynomial<R, C, R
...
-                actual: point.len(),
-            });
-        }
-
         // A trivial implementation of a polynomial evaluation at a given point.
-        let mut result = self.coeffs[0].clone();
-        for (coeff, scalar) in self.coeffs.iter().skip(1).zip(point.iter()) {
-            let term = coeff
-                .mul_by_scalar(scalar)
+        let mut result = self
```
frozenspider, line 377: I don't quite get it, why are we taking last coefficient as-is instead of a first one, and multiplying the rest? Dense polynomial stores coefficients ordered from lowers to highest degree

frozenspider: Oh sorry, I missed that you're doing `((C_{n} * X) + C_{n-1}) * X + ...`

frozenspider (approved):
Left a few minor comments, otherwise LGTM

PR #28, opened 2025-11-19
IPRS (Integer Pseudo Reed-Solomon) code

osdnk:
Implementation of a pseudo Reed-Solomon code over the integers, using a radix-NTT style recursive evaluation strategy with configurable
parameters.

In the implementation below, we made the following design choices:
- Twiddle type is generic, but must be constructible from i128 (for simplicity)
- We hardcode the (p, omega) pairs for k=1 and k=2 (assuming radix=8 and base_len=64, base_dim=32).
- We implement only the `encode` method, leaving `encode_wide` and `encode_f` unimplemented for now (as I am uncertain about their use).

The code has been tested for correctness against modular reduction for k=1 and k=2 configurations.
The code is verified against a reference implementation of NTT-based Reed-Solomon encoding (mod p).


zip-plus/src/pcs/structs.rs
```diff
@@ -17,13 +18,18 @@ pub trait ZipTypes: Send + Sync {
     /// Semiring of witness/polynomial evaluations on boolean hypercube
     type Eval: FixedSemiring + Named + ConstCoeffBitWidth;
 
+    /// Semiring of twiddle factors for FFTs
+    type Twiddle: ConstIntSemiring + ConstTranscribable + Named + Copy;
```
frozenspider: Instead of making `Twiddle` part of `ZipTypes` (as it's not relevant to RAA codes for instance), how about we make it part of IPRS codes only?
Something like
```rust
impl<Zt: ZipTypes, Tw, const REP: usize> LinearCode<Zt> for IprsCode<Zt, Tw, REP>
where
    Zt: ZipTypes,
    // Assuming we still need all these constraints
    Tw: ConstIntSemiring + ConstTranscribable + FromRef<i128> + Named + Copy,
    Zt::Cw: for<'a> MulByScalar<&'a Tw>
    IprsConfig: IprsHelpers<Tw>,
{
```
This way all the typechecking machinery will still be in place, but it will be restricted to code itself

frozenspider (changes requested):

PR #31, opened 2025-11-24
Introduce binary polynomials

ElijahVlasov:
A binary polynomial is represeneted as `u32` (degree at most 31) or `u64` (degree at most 63).

The conversion from a number to its bit polynomial is super cheap.

The projection function for binary polynomials benefits from preprocessing the powers of the random challenge as then the evaluation becomes just a sum of field elements.

Added a couple benchmarks to compare ordinary (`DensePolynomial`) and the newly inroduced binary polynomials. Evaluation of binaries is slightly faster than that of dense polynomials. Projection is massively faster.

### DIFF
--- a/poly/src/lib.rs
+++ b/poly/src/lib.rs
@@ -11,10 +11,11 @@ pub trait Polynomial<C> {
     const DEGREE_BOUND: usize;
 }
 
+// TODO: Remove the bogus type parameter `S`.
 pub trait EvaluatablePolynomial<C, S, Out>: Polynomial<C> {
     /// The type of points a polynomial can be evaluated on.
-    /// For univariate polynomials this typically is `C`,
-    /// for multivariate this is `[C]`.
+    /// For univariate polynomials this typically is `S`,
+    /// for multivariate this is `[S]`.
     type EvaluationPoint: ?Sized;
 
     /// Evaluates the polynomial at the given point.
--- a/poly/src/univariate/binary.rs
+++ b/poly/src/univariate/binary.rs
@@ -72,7 +72,7 @@ impl<T: BinaryPolyCarrier, F: PrimeField> EvaluatablePolynomial<bool, F, F> for 
             })
             .fold(F::zero_with_cfg(point.cfg()), |mut acc, coeff| {
                 acc *= point;
-                if coeff != T::ZERO { acc + &one } else { acc }
+                if !coeff.is_zero() { acc + &one } else { acc }
             }))
     }
 }
--- a/poly/src/univariate/binary.rs
+++ b/poly/src/univariate/binary.rs
@@ -1,0 +1,193 @@
+//! Dense polynomial with binary coefficients.
+use std::{fmt::Debug, ops::BitAnd};
+
+use crypto_primitives::PrimeField;
+use num_traits::ConstZero;
+use zinc_utils::projectable_to_field::ProjectableToField;
+
+use crate::{EvaluatablePolynomial, EvaluationError, Polynomial};
+
+// A type used to store the coefficients of
+// the binary polynomials. In our case, it is
+// either `u32` or `u64`.
+pub trait BinaryPolyCarrier:
+    ConstZero + TryFrom<u64, Error: Debug> + for<'a> BitAnd<&'a Self, Output = Self> + Sized + Eq
+{
+    /// The bit size of the type.
+    const BIT_SIZE: u32;
+}
+
+impl BinaryPolyCarrier for u32 {
+    const BIT_SIZE: u32 = 32;
+}
+
+impl BinaryPolyCarrier for u64 {
+    const BIT_SIZE: u32 = 64;
+}
+
+pub struct BinaryPoly<T> {
+    pub coeffs: T,
+}
+
+impl<T: BinaryPolyCarrier> Polynomial<bool> for BinaryPoly<T> {
+    const DEGREE_BOUND: usize = T::BIT_SIZE as usize;
+}
+
+impl<T: BinaryPolyCarrier> BinaryPoly<T> {
+    /// Create a polynomial of the form `X^pow_of_x`.
+    /// If `pow_of_x` exceeds the maximum degree of polynomials
+    /// of this type `None` is returned.
+    #[inline(always)]
+    pub fn single_term(pow_of_x: u32) -> Option<Self> {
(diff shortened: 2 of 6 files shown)


poly/src/lib.rs
```diff
@@ -13,8 +13,8 @@ pub trait Polynomial<C> {
 
 pub trait EvaluatablePolynomial<C, S, Out>: Polynomial<C> {
```
frozenspider, line 15: Actually, this `S` seems to be unconstrained?...

poly/src/univariate/binary.rs
```diff
@@ -0,0 +1,250 @@
...
+
+                    r_powers
+                };
+
+                move |Self { coeffs, .. }| {
+                    (0..T::BIT_SIZE)
+                        .filter(|i| {
+                            T::try_from(1 << i)
+                                .expect("If T is u32 i is 31 at most. If T is u64 i is 63 at most.")
+                                & coeffs != T::try_from(0).expect("Zero should be fine, cf. above.")
```
frozenspider: Can't we restrict `BinaryPolyCarrier: ConstZero` and use `T::ZERO` here?
If not, we should build T::try_from(0).expect(...)` in advance and just reference it in the loop.

poly/src/univariate/binary.rs
```diff
@@ -0,0 +1,250 @@
...
+            _padding: [u8; $align],
+        }
+
+        impl<T: BinaryPolyCarrier> Polynomial<bool> for $name<T> {
+            const DEGREE_BOUND: usize = T::BIT_SIZE as usize;
+        }
+
+        impl<T: BinaryPolyCarrier> $name<T> {
+            #[inline(always)]
+            pub fn x(i: u32) -> Option<Self> {
```
frozenspider: This seems to be unused and can be removed? Otherwise, the name is misleading.

poly/src/univariate/binary.rs
```diff
@@ -0,0 +1,250 @@
...
+
+// The idea of these padded binary polynomials is
+// to avoid reallocations.
+// Say, we need to turn `Vec<BinaryPoly<u32>>`
+// into `Vec<Uint<4>>` by evaluating each
+// element of the vector on a point. If we had
+// a straightforward representation of `BinaryPoly`
+// we would have to allocate a new vector and then
+// perform conversions. With padded `BinaryPoly` structures
+// we can keep everything in place.
```
frozenspider: Sorry, I don't quite get it. Is this really being used?

frozenspider:
Oh this is a DRAFT pr 🙂 I guess I shouldn't have been reviewing it yet?

poly/src/univariate/binary.rs
ElijahVlasov: I need it in the future to define rotation on binary polys, i.e. multiplication by `X ^ i` mod `X^32 - 1`.

Would it be better to call it `pow_of_x`?

poly/src/univariate/binary.rs
ElijahVlasov: cf. https://github.com/NethermindEth/zinc-plus/blob/7a0bbac32a4a14df5df829651ac46855b583361a/poly/src/univariate/binary.rs#L274 from the next pr.

It's handy to prepare a vector of binary polys for a sumcheck mod random prime. This way we don't need to reallocate a really big vector of binary polys.

poly/src/lib.rs
ElijahVlasov: Yeah, it is. I tried removing it but then I failed to get the traits right in `ZipTypes`. I can add a todo for now and have a more careful look once more later.

poly/src/univariate/binary.rs
ElijahVlasov: And it's done :grin:

poly/src/lib.rs
ElijahVlasov: And it's done :grimacing:

poly/src/univariate/binary.rs
```diff
@@ -0,0 +1,245 @@
...
+                    r_powers
+                };
+
+                move |Self { coeffs, .. }| {
+                    (0..T::BIT_SIZE)
+                        .filter(|i| {
+                            T::try_from(1 << i)
+                                .expect("If T is u32 i is 31 at most. If T is u64 i is 63 at most.")
+                                & coeffs
+                                != T::ZERO
```
frozenspider: My bad, this should've been `!coeffs.is_zero()`

poly/src/univariate/binary.rs
```diff
@@ -0,0 +1,245 @@
...
+            _padding: [u8; $align],
+        }
+
+        impl<T: BinaryPolyCarrier> Polynomial<bool> for $name<T> {
+            const DEGREE_BOUND: usize = T::BIT_SIZE as usize;
+        }
+
+        impl<T: BinaryPolyCarrier> $name<T> {
+            #[inline(always)]
+            pub fn pow_of_x(i: u32) -> Option<Self> {
```
frozenspider: From the code I can deduce that it creates a poly with only `x^i` term present - however, I couldn't have surmised that from the method name. Maybe call it `new_single_term(pow_of_x: u32)` and/or add a descriptive rustdoc comment?

poly/src/univariate/binary.rs
frozenspider: I think either #32 should be merged with this one, or this should be moved to #32 - really hard to review them separately

ElijahVlasov:
Reworked this PR entirely. Now these are sane binary polynomials.

poly/src/univariate/binary.rs
```diff
@@ -0,0 +1,193 @@
...
+        let one = F::one_with_cfg(point.cfg());
+        Ok((0..T::BIT_SIZE)
+            .map(|i| {
+                T::try_from(1u64 << ((T::BIT_SIZE - 1) - i))
+                    .expect("T is either u32 or u64 and this should always go through.")
+                    & &self.coeffs
+            })
+            .fold(F::zero_with_cfg(point.cfg()), |mut acc, coeff| {
+                acc *= point;
+                if coeff != T::ZERO { acc + &one } else { acc }
```
frozenspider: Nit: `!coeff.is_zero()`

frozenspider (approved):
Looks much cleaner now, great job!

PR #32, opened 2025-11-24
Unsafe projection onto `Field::Inner`

ElijahVlasov:
When projecting a lot of elements onto a field (think of projecting a `DenseMultilinearExtension<BinaryPoly<u32>>`) it gets quite expensive to drag a field config with each. Instead we can project onto the field and only store the inner Montgomery representation. For that we introduce the unsafe trait:
```rust
pub unsafe trait ProjectableToFieldInner<F: PrimeField> {
    fn prepare_projection_to_inner(sampled_value: &F) -> impl Fn(*mut Self) + 'static;
}
```
The returned closure is supposed to perform the transformation above and store the result in the place.

We also define this utilitary function:
```rust
pub unsafe fn project_vec_in_place<T, F, P>(mut vec: Vec<T>, project: P) -> Vec<F>
where
    T: Send + Sync,
    F: Send + Sync,
    P: Fn(*mut T) + Send + Sync,
```
that can be used to batch-project vectors.

### DIFF
--- a/poly/src/univariate/binary.rs
+++ b/poly/src/univariate/binary.rs
@@ -1,8 +1,8 @@
 //! Dense polynomial with binary coefficients.
 use std::{fmt::Debug, ops::BitAnd};
 
-use crypto_primitives::PrimeField;
-use zinc_utils::projectable_to_field::ProjectableToField;
+use crypto_primitives::{Field, PrimeField, crypto_bigint_monty::MontyField};
+use zinc_utils::projectable_to_field::{ProjectableToField, ProjectableToFieldInner};
 
 use crate::{EvaluatablePolynomial, EvaluationError, Polynomial};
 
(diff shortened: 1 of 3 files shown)


poly/src/univariate/binary.rs
```diff
@@ -157,16 +157,38 @@ define_binary_poly!(BinaryPoly64, 0);
...
 
+macro_rules! impl_projectable_to_field_inner {
+    ($for:ident, $field:ty) => {
+        unsafe impl<T: BinaryPolyCarrier> ProjectableToFieldInner<$field> for $for<T> {
+            fn prepare_projection_to_inner(sampled_value: &$field) -> impl Fn(*mut Self) + 'static {
+                let project = Self::prepare_projection(sampled_value);
+                move |p| unsafe {
+                    let val = project(p.as_ref().expect("A correct pointer was expected here"));
+                    let p = p as *mut <$field as Field>::Inner;
+                    *p = val.into_inner();
```
frozenspider, line 168: 1. This operation is VERY unsafe. There's no guarantees that `BinaryPolyXxx<T>` has exactly the same memory layout as `Field::Inner`. To do this "safely" would require a lot of compile-time/runtime checks (similarly to [what ReinterpretVector did](https://github.com/NethermindEth/zinc-plus/blob/cab02cedf7d9ca5975961093e88fb1b6e9bb0221/zip-plus/src/utils.rs#L358-L373)), and it would still be unsafe and easy to break with any kind of change.
2. I don't get what problem you're solving this way. If you simply don't want to drag config along, why not project polynomial into `Field::Inner` and be done with it? So, something like
```rust
pub trait ProjectableToFieldInner<F: PrimeField> {
    fn prepare_projection_to_inner(sampled_value: &F) -> impl Fn(&Self) -> F::Inner + 'static;
}

impl<T: BinaryPolyCarrier> ProjectableToFieldInner<MontyField<1>> for BinaryPoly64<T> {
    fn prepare_projection_to_inner(sampled_value: &MontyField<1>) -> impl Fn(&Self) -> <MontyField::<1> as Field>::Inner + 'static {
        let project = Self::prepare_projection(sampled_value);
        move |p| {
            project(p).into_inner()
        }
    }
}
```
This code is completely safe.
Do you see problems with this approach?

utils/src/projectable_to_field.rs
```diff
@@ -7,3 +10,55 @@ pub trait ProjectableToField<F: PrimeField> {
...
+    F: Send + Sync,
+    P: Fn(*mut T) + Send + Sync,
+{
+    assert_eq!(size_of::<T>(), size_of::<F>());
+
+    cfg_iter_mut!(vec).for_each(|x| {
+        project(x);
+    });
+
+    unsafe { std::mem::transmute(vec) }
```
frozenspider, line 49: `transmute` is [unsafe even in the unsafe land](https://doc.rust-lang.org/nomicon/transmutes.html), there should be an absolutely compelling reason to use it. Even in this case this should've been `Vec::from_raw_parts` - however, I'm not sure this is needed at all.
Are you sure than this is such a big win in performance compared to `iter().map().collect_vec()` not in place?

frozenspider (changes requested):

ElijahVlasov:
No, this doesn't seem to have any benefits against safe version! Ran a benchmark. Nor padding the binary poly! I'll rework it a bit.

PR #33, opened 2025-11-24
Natural evaluation domain polynomial

ElijahVlasov:
Now sumcheck prover and verifier exchange `NatEvaluationPoly` messages.

The type implements `EvaluatablePolynomial` trait.

I hope to abstract away the entire univariate logic from sumcheck one day to be able to swap painlessly between evaluation domain and coefficient domain to see which one is faster. This is a first step towards that.


piop/src/sumcheck/prover.rs
```diff
@@ -57,7 +53,7 @@ where
         v_msg: &Option<F>,
         comb_fn: impl Fn(&[F]) -> F + Send + Sync,
         config: &F::Config,
-    ) -> ProverMsg<F> {
+    ) -> NatEvaluatedPoly<F> {
```
frozenspider: I think it still makes sense to keep it wrapped in `ProverMsg`, for added type safety and verbosity. Could be a simple newtype `pub struct ProverMsg<F>(pub NatEvaluatedPoly<F>)`

poly/src/univariate/nat_evaluation.rs
```diff
@@ -0,0 +1,179 @@
...
+    /// Interpolate the *unique* univariate polynomial of degree *at most*
+    /// `evaluations.len()-1` passing through the y-values in `evaluations` at x
+    /// = 0,..., evaluations.len()-1
+    /// and evaluate this  polynomial at `point`. In other words, efficiently
+    /// compute  $\sum_{i=0}^{len\ evaluations - 1} evaluations\[i\] *
+    /// (\prod_{j!=i} (\text{point} - j)/(i-j))$.
+    // All the arithmetic ops in the function
+    // are made sure to not overflow.
+    #[allow(clippy::arithmetic_side_effects, clippy::cast_possible_wrap)]
+    #[allow(clippy::arithmetic_side_effects, clippy::cast_possible_wrap)]
```
frozenspider: Double-annotation

piop/src/sumcheck/verifier.rs
```diff
@@ -117,7 +117,7 @@ impl<F: FromPrimitiveWithConfig> VerifierState<F> {
             panic!("insufficient rounds");
         }
         for i in 0..self.nv {
-            let evaluations = &self.polynomials_received[i];
+            let NatEvaluatedPoly { evaluations } = &self.polynomials_received[i];
```
frozenspider: Nit: destructuring assignment is more verbose and not really needed with one field only, but it's not a big deal

ElijahVlasov: That was the cheapest way in terms of the number of diff lines :grin:

frozenspider: Is it though? 😄 
```diff
- let evaluations = &self.polynomials_received[i];
+ let evaluations = &self.polynomials_received[i].evaluations;
```

piop/src/sumcheck/prover.rs
```diff
@@ -4,14 +4,26 @@ use ark_std::{cfg_into_iter, cfg_iter_mut, slice, vec, vec::Vec};
...
-    pub(crate) evaluations: Vec<F>,
+pub struct ProverMsg<F>(pub NatEvaluatedPoly<F>);
+
+impl<F> From<NatEvaluatedPoly<F>> for ProverMsg<F> {
+    fn from(polynomial: NatEvaluatedPoly<F>) -> Self {
+        Self(polynomial)
+    }
+}
+
+impl<F> From<ProverMsg<F>> for NatEvaluatedPoly<F> {
```
frozenspider: I think these unnecessarily obscure what's going on - and hurt type safety to some extent.
You can always do
```rust
ProverMsg(poly) // wrap
prover_msg.0 // unwrap
```
and it's not even verbose

frozenspider (approved):

piop/src/sumcheck/verifier.rs
ElijahVlasov: You're right!

PR #35, opened 2025-11-27
Remove the dummy `S` type parameter from `EvaluatablePolynomial`

ElijahVlasov:
It seems to not be used anywhere and removing it breaks nothing 🤸🏻‍♂️

### DIFF
--- a/poly/src/univariate/dense.rs
+++ b/poly/src/univariate/dense.rs
@@ -385,17 +385,14 @@ impl<R: Semiring, const DEGREE_PLUS_ONE: usize> Polynomial<R>
     const DEGREE_BOUND: usize = DEGREE_PLUS_ONE - 1;
 }
 
-impl<R: Semiring, C, const DEGREE_PLUS_ONE: usize> EvaluatablePolynomial<R, C, R>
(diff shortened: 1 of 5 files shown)


poly/src/univariate/dense.rs
```diff
@@ -385,17 +385,14 @@ impl<R: Semiring, const DEGREE_PLUS_ONE: usize> Polynomial<R>
     const DEGREE_BOUND: usize = DEGREE_PLUS_ONE - 1;
 }
 
-impl<R: Semiring, C, const DEGREE_PLUS_ONE: usize> EvaluatablePolynomial<R, C, R>
+impl<R: Semiring, const DEGREE_PLUS_ONE: usize> EvaluatablePolynomial<R, R>
     for DensePolynomial<R, DEGREE_PLUS_ONE>
 where
-    R: for<'a> MulByScalar<&'a C>,
+    R: for<'a> MulByScalar<&'a R>,
```
frozenspider: If this is all we want, do we really need this constraint? R is already `Mul<Rhs=&R>`

ElijahVlasov: It seems so. Neither tests nor benches ever use it.

frozenspider (approved):

PR #39, opened 2025-11-28
Lift `Output` in `InnerProduct` to generic parameters

ElijahVlasov:
* Now it's possible to define inner products with different resulting types for the same input arguments. It's convenient when we want to enlarge the bit width of the result (e.g. `i128` -> `Int<M>`).

* IMPORTANT: this PR also increments the number of `alphas` drawn at the testing phase. Previously, we would take the linear combination `1 * p_0 + \alpha_0 * p_1 + ... + \alpha_{DEGREE_BOUND - 1} * p_{DEGREE_BOUND}`. Now we do `\alpha_0 * p_0 + \alpha_1 * p_1 + ... + \alpha_{DEGREE_BOUND} * p_{DEGREE_BOUND}`.

### DIFF
--- a/utils/src/inner_product.rs
+++ b/utils/src/inner_product.rs
@@ -1,20 +1,13 @@
-use crypto_primitives::crypto_bigint_int::Int;
+use crypto_primitives::{boolean::Boolean, crypto_bigint_int::Int};
 use num_traits::CheckedAdd;
 use thiserror::Error;
 
-use crate::mul_by_scalar::MulByScalar;
+use crate::{from_ref::FromRef, mul_by_scalar::MulByScalar};
 
 /// A trait for types that can be dot-multiplied.
-pub trait InnerProduct<Rhs> {
-    /// The resulting type we get after the inner product.
-    type Output;
-
+pub trait InnerProduct<Rhs, Output> {
     /// The main entry point for the inner product.
-    fn inner_product(
-        &self,
-        rhs: &[Rhs],
-        zero: Self::Output,
-    ) -> Result<Self::Output, InnerProductError>;
+    fn inner_product(&self, rhs: &[Rhs], zero: Output) -> Result<Output, InnerProductError>;
 }
 
 #[derive(Clone, Debug, PartialEq, Error)]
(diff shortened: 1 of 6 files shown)


ElijahVlasov:
> * IMPORTANT: this PR also increments the number of `alphas` drawn at the testing phase. Previously, we would take the linear combination `1 * p_0 + \alpha_0 * p_1 + ... + \alpha_{DEGREE_BOUND - 1} * p_{DEGREE_BOUND}`. Now we do `\alpha_0 * p_0 + \alpha_1 * p_1 + ... + \alpha_{DEGREE_BOUND} * p_{DEGREE_BOUND}`.

REVIEW THIS CAREFULLY

frozenspider:
> * IMPORTANT: this PR also increments the number of `alphas` drawn at the testing phase. Previously, we would take the linear combination `1 * p_0 + \alpha_0 * p_1 + ... + \alpha_{DEGREE_BOUND - 1} * p_{DEGREE_BOUND}`. Now we do `\alpha_0 * p_0 + \alpha_1 * p_1 + ... + \alpha_{DEGREE_BOUND} * p_{DEGREE_BOUND}`.

Does this mean there's no more constant term (x^0) in a polynomial, and primitives are no longer considered constant terms?
Or we're changing the semantics of what inner product does for those types?

ElijahVlasov:
> > * IMPORTANT: this PR also increments the number of `alphas` drawn at the testing phase. Previously, we would take the linear combination `1 * p_0 + \alpha_0 * p_1 + ... + \alpha_{DEGREE_BOUND - 1} * p_{DEGREE_BOUND}`. Now we do `\alpha_0 * p_0 + \alpha_1 * p_1 + ... + \alpha_{DEGREE_BOUND} * p_{DEGREE_BOUND}`.
> 
> Does this mean there's no more constant term (x^0) in a polynomial, and primitives are no longer considered constant terms? Or we're changing the semantics of what inner product does for those types?

2025-12-02-111313_hyprshot

The intended semantics of this inner product in the paper is `\alpha_0 * p_0 + \alpha_1 * p_1 + ... + \alpha_{DEGREE_BOUND} * p_{DEGREE_BOUND}`. We could take \alpha_0 = 1 and that'd be fine. And that's what we're doing in the code currently. To simplify things a bit and not to make a weird implementation of inner product where we ignore the first element of rhs or assume it is `1` I suggest we do it the proper way.  

I suggest we implement inner product for scalar types as multiplication, e.g. `(x: Int<M>).inner_product([y], ...) = x * y`. However, in the testing phase we check if the degree bound is `0` we don't sample any challenges and put `\alphas=[1]`.

utils/src/inner_product.rs
```diff
@@ -25,65 +18,83 @@ pub enum InnerProductError {
...
+impl<Lhs, Rhs, Out> InnerProduct<Rhs, Out> for &[Lhs]
 where
-    Lhs: Clone + CheckedAdd + for<'z> MulByScalar<&'z Rhs>,
+    Lhs: CheckedAdd + for<'a> MulByScalar<&'a Rhs>,
+    Out: From<Lhs> + CheckedAdd,
 {
-    type Output = Lhs;
-
-    fn inner_product(&self, rhs: &[Rhs], zero: Lhs) -> Result<Self::Output, InnerProductError> {
+    fn inner_product(&self, rhs: &[Rhs], zero: Out) -> Result<Out, InnerProductError> {
```
frozenspider, line 26: This `if self.len() != rhs.len() {` check should be added here as well

utils/src/inner_product.rs
```diff
@@ -25,65 +18,83 @@ pub enum InnerProductError {
...
+            Ok(Out::from_ref(
+                &self
+                    .mul_by_scalar(&point[0])
+                    .ok_or(InnerProductError::Overflow)?,
+            ))
         }
     }
 }
 
+impl<const M: usize> InnerProduct<i128, Int<M>> for &[Boolean] {
```
frozenspider: Too bad we can't make it fully generic (in stable Rust anyway), but we could at least do
```rust
impl<Out> InnerProduct<i128, Out> for &[Boolean]
where
    Out: CheckedAdd + From<i128>,
```
to not be bound to `Int<M>`

frozenspider (approved):

PR #40, opened 2025-11-28
Test phase optimisation

ElijahVlasov:
* Remove `.map(Zt::Comb::from_ref)` in
```rust
let evals: Vec<_> = poly
                .evaluations
                .iter()
                .map(Zt::Comb::from_ref)
                .map(|p| p.inner_product(&alphas, Zt::CombR::ZERO))
                .try_collect()?;
```
that regresses performance of the test phase. In the main benchmark, `Zt::Comb` is `DensePolynomial<Int<5>, D_PLUS_ONE>` and `Zt::Eval` is `DensePolynomial<Boolean, D_PLUS_ONE>`. So the conversion would erase the information that the coefficients of the polynomial are 0s or 1s, and a less optimal `inner_product` implementation would be used.

* Optimise trait constraints for `Eval` of `ZipTypes`. Now it's the bare minimum of what we need to know about `Eval`.

### DIFF
--- a/utils/src/mul_by_scalar.rs
+++ b/utils/src/mul_by_scalar.rs
@@ -1,6 +1,6 @@
 use crate::from_ref::FromRef;
 use crypto_primitives::{boolean::Boolean, crypto_bigint_int::Int};
-use num_traits::CheckedMul;
+use num_traits::{CheckedMul, ConstZero};
 
 pub trait MulByScalar<Rhs>: Sized {
     /// Multiplies the current element by a scalar from the right (usually - a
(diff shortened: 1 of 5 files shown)


ElijahVlasov:
70-80% improvement for Test benches

utils/src/mul_by_scalar.rs
```diff
@@ -45,3 +45,9 @@ macro_rules! impl_mul_int_by_primitive_scalar {
 }
 
 impl_mul_int_by_primitive_scalar!(i8, i16, i32, i64, i128);
+
+impl MulByScalar<&Boolean> for i128 {
```
frozenspider: I think it makes sense to blanket-implement it for any `T: Clone`

frozenspider (approved):

utils/src/mul_by_scalar.rs
ElijahVlasov: That conflicts with this unfortunately :cry:  https://github.com/NethermindEth/zinc-plus/blob/c07d8c989803574115f2688f8acbc70611fe523a/poly/src/univariate/dense.rs#L485

frozenspider: This is somewhat of a hacky solution, but...
```rust
impl<T> MulByScalar<&Boolean> for T where T: Clone + ConstZero + From<i8> {
    fn mul_by_scalar(&self, rhs: &Boolean) -> Option<Self> {
        Some(if rhs.into_inner() { self.clone() } else { ConstZero::ZERO })
    }
}
```
This covers all signed primitives as well as `Int<_>`

frozenspider: Could even be `From<bool>` to cover `Boolean` itself

PR #45, opened 2025-12-02
BoxedMontyField -> MontyField

ElijahVlasov:


utils/src/from_ref.rs
```diff
@@ -80,3 +80,10 @@ impl<const LIMBS: usize, const LIMBS2: usize> FromRef<Int<LIMBS2>> for Int<LIMBS
         Self::try_from(value.inner()).expect("Destination Int type is too small")
     }
 }
+
+impl<const LIMBS: usize, const LIMBS2: usize> FromRef<Uint<LIMBS2>> for Uint<LIMBS> {
+    #[inline]
+    fn from_ref(value: &Uint<LIMBS2>) -> Self {
+        Self::try_from(value.inner()).expect("Destination Int type is too small")
```
frozenspider: Typo, should be `Uint type` - also not sure if this piece is needed for the benchmark change

frozenspider (approved):

utils/src/from_ref.rs
ElijahVlasov: It is! Some trait constraints weren't happy

PR #46, opened 2025-12-03
Remove binary polynomials

ElijahVlasov:
With introduction of the `InnerProductProjection` there's not much point in having `BinaryPoly` type. It's projection falls slightly behind that of `DensePoly<Boolean, _>`. They're probably better memory wise, but it's not great of advantage. So, removing them.


ElijahVlasov:
Overall, binary polynomials were a premature optimisation but they showed what can be done better.

frozenspider (approved):

PR #48, opened 2025-12-05
Compile time configuration for linear codes

ElijahVlasov:
Since we always should know the configuration of the codes at compile time there is no need to drag them along in runtime.

It does bring some nuisance in tests but that's addressed via macros.

### DIFF
--- a/zip-plus/benches/zip_plus_benches.rs
+++ b/zip-plus/benches/zip_plus_benches.rs
@@ -52,8 +52,8 @@ impl<const D_PLUS_ONE: usize> ZipTypes for BenchZipPlusTypes<D_PLUS_ONE> {
 #[derive(Clone, Copy)]
 struct BenchRaaConfig;
 impl RaaConfig for BenchRaaConfig {
-    const PERMUTE_IN_PLACE: bool = false;
-    const CHECK_FOR_OVERFLOWS: bool = true;
+    const PERMUTE_IN_PLACE: bool = true;
+    const CHECK_FOR_OVERFLOWS: bool = false;
 }
 
 type Code<const D_PLUS_ONE: usize> = RaaCode<BenchZipPlusTypes<D_PLUS_ONE>, BenchRaaConfig, 4>;
--- a/zip-plus/benches/zip_plus_benches.rs
+++ b/zip-plus/benches/zip_plus_benches.rs
@@ -16,7 +16,7 @@ use criterion::{Criterion, criterion_group, criterion_main};
 use crypto_bigint::U64;
 use crypto_primitives::{boolean::Boolean, crypto_bigint_int::Int, crypto_bigint_uint::Uint};
 use zip_plus::{
-    code::raa::{RaaCode, RaaConfigGeneric},
+    code::raa::{RaaCode, RaaConfig},
     pcs::structs::ZipTypes,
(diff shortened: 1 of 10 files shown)


ElijahVlasov:
The commit [201b968](https://github.com/NethermindEth/zinc-plus/pull/48/commits/201b96898ccc77ad1e963be7f13c751e10ed412d) provides a pedestrian version of the type level RAA config - we introduce constants
```rust
pub const PERMUTE_IN_PLACE: bool = true;
pub const DONT_PERMUTE_IN_PLACE: bool = false;

pub const CHECK_FOR_OVERFLOWS: bool = true;
pub const DONT_CHECK_FOR_OVERFLOWS: bool = false;
``` 
that can be used to fill in the config at compile time.

[03fbb01](https://github.com/NethermindEth/zinc-plus/pull/48/commits/03fbb016e46c04284a14be7b42cd8ce1560cd8be) is a more of an over the top type level yoga.

Choose the one that you like more :slightly_smiling_face:

zip-plus/src/code/raa.rs
```diff
@@ -5,12 +5,36 @@ use std::{marker::PhantomData, ops::AddAssign};
 use zinc_poly::ConstCoeffBitWidth;
 use zinc_utils::{add, from_ref::FromRef, mul};
 
+pub trait RaaConfig: Copy + Send + Sync {
+    const PERMUTE_IN_PLACE: bool;
+    const CHECK_FOR_OVERFLOWS: bool;
+}
+
+impl RaaConfig for () {
```
frozenspider: Hmm. Instead of providing implementations here, how about we define it at the caller level (i.e. test utils and benches)? That would resemble the previous approach where RAA itself won't care about setting the defaults, leaving that up to callers -- well, "definers" in this case, but you got the point 🙂

zip-plus/benches/zip_plus_benches.rs
```diff
@@ -49,12 +49,14 @@ impl<const D_PLUS_ONE: usize> ZipTypes for BenchZipPlusTypes<D_PLUS_ONE> {
     >;
 }
 
-type Code<const D_PLUS_ONE: usize> = RaaCode<BenchZipPlusTypes<D_PLUS_ONE>, 4>;
+#[derive(Clone, Copy)]
+struct BenchRaaConfig;
+impl RaaConfig for BenchRaaConfig {
+    const PERMUTE_IN_PLACE: bool = false;
+    const CHECK_FOR_OVERFLOWS: bool = true;
```
frozenspider: Note that this mismatches previous config! Is this intended?

zip-plus/src/code/raa.rs
```diff
@@ -24,20 +27,10 @@ pub struct RaaCode<Zt: ZipTypes, const REP: usize> {
...
 
-    phantom: PhantomData<Zt>,
-}
-
-#[derive(Debug, Clone, Copy)]
-pub struct RaaConfig {
-    /// Whether to check for overflows during encoding
-    pub check_for_overflows: bool,
-
-    /// Whether to permute the codeword in place, instead of copying it using a
```
frozenspider, line 35: I think these comments should be preserved

frozenspider (approved):

PR #49, opened 2025-12-12
Avoid vectors in checked operations for `DensePolynomial`

ElijahVlasov:
Avoid extra manipulations with vectors in the `CheckedAdd`, `CheckedSub`, `CheckedNeg`, and `MulByScalar` implementations for `DensePolynomial`

### DIFF
--- a/poly/src/univariate/dense.rs
+++ b/poly/src/univariate/dense.rs
@@ -272,40 +272,47 @@ impl<'a, R: Semiring, const DEGREE_PLUS_ONE: usize> MulAssign<&'a Self>
     }
 }
 
-impl<R: Ring, const DEGREE_PLUS_ONE: usize> CheckedNeg for DensePolynomial<R, DEGREE_PLUS_ONE> {
+impl<R: Ring + Zero, const DEGREE_PLUS_ONE: usize> CheckedNeg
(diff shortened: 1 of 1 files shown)


frozenspider (commented):
Have you benchmarked old vs new code? How much are we winning?

ElijahVlasov:
> Have you benchmarked old vs new code? How much are we winning?

We're winning quite a lot in the IPRS PR thanks to this :slightly_smiling_face: 

It was like ~15%

frozenspider (approved):
My bet would be that the performance win comes from `coeff.is_zero()` check alone, not the change in coeffs cloning part. etc. 🙂

ElijahVlasov:
> My bet would be that the performance win comes from `coeff.is_zero()` check alone, not the change in coeffs cloning part. etc. 🙂

Yeah, you must be right. Tho, the code is slightly cleaner I'd argue

ElijahVlasov:
> > My bet would be that the performance win comes from `coeff.is_zero()` check alone, not the change in coeffs cloning part. etc. 🙂
> 
> Yeah, you must be right. Tho, the code is slightly cleaner I'd argue

In my taste :upside_down_face:

PR #50, opened 2025-12-12
Introduce a new trait `CheckedAddAssign` for in-place checked addition

ElijahVlasov:


frozenspider (commented):
Sorry, I don't quite get why is this needed in the first place.
1. This doesn't seem to be used much in the current codebase.
2. We can always do `x = x.checked_add(y)` without any additional code, and that's not longer/harder than `x.checked_add_assign(y)`.

ElijahVlasov:
> Sorry, I don't quite get why is this needed in the first place.
> 
> 1. This doesn't seem to be used much in the current codebase.
> 2. We can always do `x = x.checked_add(y)` without any additional code, and that's not longer/harder than `x.checked_add_assign(y)`.

Yeah, not much of benefit for IPRS either

PR #51, opened 2025-12-12
IPRS codes implementation

ElijahVlasov:
This PR introduces a radix-8 pseudo number theoretic transform and IPRS codes based on it.


poly/src/univariate/dense.rs
```diff
@@ -617,3 +611,17 @@ where
...
+        coeffs[0] = value;
+        Self { coeffs }
+    }
+}
+
+impl<const DEGREE_PLUS_ONE: usize> FromRef<i64> for DensePolynomial<i128, DEGREE_PLUS_ONE> {
+    fn from_ref(value: &i64) -> Self {
+        Self::from(i128::from(*value))
+    }
+}
```
frozenspider: Please keep `From`/`FromRef` (and all other traits, for that matters) implementations grouped together in the file. This makes the code easier to maintain.

zip-plus/src/code/iprs/pntt/radix8/butterfly.rs
```diff
@@ -0,0 +1,72 @@
...
+            2 => generate_radix_8_butterfly!(2),
+            3 => generate_radix_8_butterfly!(3),
+            4 => generate_radix_8_butterfly!(4),
+            5 => generate_radix_8_butterfly!(5),
+            6 => generate_radix_8_butterfly!(6),
+            7 => generate_radix_8_butterfly!(7),
+            _ => panic!("Incorrect butterfly number"),
+        }
+    };
+}
```
frozenspider: If this is just 8 * 8 table, how about we generate it at compile time rather than at runtime?
```suggestion
const BUTTERFLY: [[usize; 8]; 8] = {
    macro_rules! generate_radix_8_butterfly {
        ($i:expr) => {
            [
                0,
                $i % 8,
                (2 * $i) % 8,
                (3 * $i) % 8,
                (4 * $i) % 8,
                (5 * $i) % 8,
                (6 * $i) % 8,
                (7 * $i) % 8,
            ]
        };
    }
    let mut result = [[0; 8]; 8];
    const_for!((i in 0..(result.len())) {
        result[i] = generate_radix_8_butterfly!(i);
    });
    result
};
```

zip-plus/src/code/iprs/pntt/radix8/butterfly.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// Apply butterfly given by `twiddles` to a slice
+/// of subresults in `x`. Use `mul_by_twiddle` as a means
+/// to multiply `Out` by `Twiddle`.
+pub(crate) fn radix_8_butterfly<Out, Twiddle, M, const I: usize>(
```
frozenspider: This doesn't seem to be used outside of `do_all_butterflies` - so how about we make this private, and transform `do_all_butterflies` into a public function?

zip-plus/src/code/iprs/pntt/radix8/params.rs
```diff
@@ -0,0 +1,152 @@
...
+    const BASE_DIM: usize;
+    /// The coefficients used to combine subresults.
+    /// They are the 8-th roots of unity from the field `Self::Field`
+    /// lifted to `Self::Int`.
+    const TWIDDLES: [Self::Int; 8];
+
+    /// The length of the pseudo NTT's input.
+    const N: usize = Self::BASE_LEN * (1 << (3 * Self::DEPTH));
+    /// The length of the pseudo NTT's output.
+    const M: usize = Self::BASE_DIM * (1 << (3 * Self::DEPTH));
```
frozenspider: Nit: Call them `INPUT_LEN/OUTPUT_LEN`?

zip-plus/src/code/iprs/pntt/radix8/params.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    /// A helper to get an integer from a field element in
+    /// the range `[-(p - 1)/2, (p - 1)/2]`.
```
frozenspider: Slightly misleading comment reads as if a field element itself is in the said range

zip-plus/src/code/iprs/pntt/radix8/params.rs
```diff
@@ -0,0 +1,152 @@
...
+    fn field_to_int_normalized(x: Self::Field) -> Self::Int;
+}
+
+#[derive(Clone, Debug)]
+// Precomputed parameters needed for
+// pseudo NTT algorithm.
```
frozenspider, line 129: Use rustdoc before the `#[derive]`

zip-plus/src/code/iprs/pntt/radix8/params.rs
```diff
@@ -0,0 +1,152 @@
...
+pub struct Radix8PNTTParams<C: Config> {
```
frozenspider: As per [Rust naming conventions](https://rust-lang.github.io/api-guidelines/naming.html), acronyms and contractions of compound words count as one word in `UpperCamelCase`

zip-plus/src/code/iprs/pntt/radix8/params.rs
```diff
@@ -0,0 +1,152 @@
...
+    /// The base matrix of the pseudo NTT.
+    pub base_matrix: Vec<Vec<C::Int>>,
```
frozenspider: Maybe use `DenseRowMatrix` for this?
(Ideally this should've been `[[C::Int; C::BASE_LEN]; C::BASE_DIM]` but const generics are not stabilized yet)

zip-plus/src/code/iprs/pntt/radix8/params.rs
```diff
@@ -0,0 +1,152 @@
...
+            .map(C::field_to_int_normalized)
+            .collect_vec()
+    }
+}
+
+#[derive(Clone, Copy)]
+/// Pseudo NTT configuration derived from
+/// the field `Fp` for `p = 2^16 + 1`.
+///
+/// Supports `DEPTH` up to `3`.
```
frozenspider, line 161: Ditto about order of `#[annotations]` and rustdocs, and naming conventions for acronyms

zip-plus/src/code/iprs/pntt/radix8/params.rs
```diff
@@ -0,0 +1,152 @@
...
+        let big_int = MontBackend::<FqConfig, 1>::into_bigint(x);
+
+        normalize_field_element(big_int.0[0], (1 << 16) + 1)
+    }
+}
+
+#[allow(clippy::arithmetic_side_effects)]
+#[allow(clippy::cast_possible_wrap)]
+/// Field normalization for at most 32-bit fields.
+/// Might have unpleasant overflows if used for bigger fields.
```
frozenspider: Ditto: comment/annotations order
(There are many other places like this, not marking them further)

zip-plus/src/code/iprs/pntt/radix8/params.rs
```diff
@@ -0,0 +1,152 @@
...
+    type Field = Fq;
+    const BASE_LEN: usize = 32;
+    const BASE_DIM: usize = 64;
+    const DEPTH: usize = DEPTH;
+    const TWIDDLES: [Self::Int; 8] = [1, 4096, -256, 16, -1, -4096, 256, -16];
+
+    fn field_to_int_normalized(x: Self::Field) -> Self::Int {
+        let big_int = MontBackend::<FqConfig, 1>::into_bigint(x);
+
+        normalize_field_element(big_int.0[0], (1 << 16) + 1)
```
frozenspider: Maybe define `const MODULUS` in `mod fq`?

zip-plus/src/code/iprs/pntt/radix8/params.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+#[derive(Clone, Copy)]
+/// Pseudo NTT configuration derived from
+/// the field `Fp` for `p = 2^16 + 1`.
+///
+/// Supports `DEPTH` up to `3`.
+pub struct PNTTConfigF2_16_1<const DEPTH: usize>;
+
+pub(crate) mod fq {
```
frozenspider: This doesn't need to be `pub(crate)`

zip-plus/src/code/iprs/pntt/radix8/params.rs
```diff
@@ -0,0 +1,152 @@
...
+    #![allow(non_local_definitions)]
+    use ark_ff::{Fp64, MontBackend, MontConfig};
+    #[derive(MontConfig)]
+    #[modulus = "65537"]
+    #[generator = "3"]
+    pub struct FqConfig;
+    pub type Fq = Fp64<MontBackend<FqConfig, 1>>;
```
frozenspider: I think it makes sense to define `MontBackend<FqConfig, 1>` as its own type alias, given that it's also used in `field_to_int_normalized`

zip-plus/src/code/iprs/pntt/radix8/params.rs
```diff
@@ -0,0 +1,152 @@
...
+    use super::*;
+
+    // Twiddles are indeed the 8th roots of unity.
+    fn check_twiddles_generic<C: Config>() {
+        let domain = Radix2EvaluationDomain::<C::Field>::new(8).unwrap();
+
+        let expected = domain
+            .elements()
+            .map(C::field_to_int_normalized)
+            .collect_vec();
```
frozenspider: Since that's exactly what we're doing in `precompute_roots_of_unity`, maybe reuse it here?

zip-plus/src/code/iprs.rs
```diff
@@ -0,0 +1,157 @@
+mod pntt;
+
+use num_traits::CheckedAdd;
+use std::{iter::Sum, marker::PhantomData, ops::AddAssign};
+
+use crypto_primitives::{FromPrimitiveWithConfig, FromWithConfig};
+use itertools::Itertools;
+use num_traits::{CheckedMul, ConstZero, Zero};
```
frozenspider: rustfmt does not group `use`s separated by an empty line. That's why there's two `use num_traits::whatever`

zip-plus/src/code/iprs/pntt/radix8.rs
```diff
@@ -0,0 +1,261 @@
...
+use rayon::prelude::*;
+use std::{array, iter::Sum};
+use zinc_utils::add;
+
+use butterfly::*;
+use octet_reversal::*;
+
+pub(crate) use mul_by_twiddle::*;
+
+pub use params::*;
```
frozenspider: If we're reexporting `params`, why not just make it a public module?

zip-plus/src/code/iprs.rs
```diff
@@ -0,0 +1,157 @@
...
+        Out: From<In>
+            + Clone
+            + CheckedAdd
+            + for<'a> AddAssign<&'a Out>
+            + CheckedMul
+            + for<'a> MulByScalar<&'a Config::Int>
+            + Send
+            + Sync
+            + Sum,
+        for<'a> &'a Out: From<&'a In>,
```
frozenspider: This seems like a really weird requirement, `In` being reinterpretable as `Out`, it's pretty much never going to work for distinct `In` and `Out`.
In practice, it's used in just two places - `encode` and `encode_wide`. In both those cases you first coerce `In` and `Out` to be the exact same type.
Given that, type signature should be simplified to have just one type.

zip-plus/src/code/iprs/pntt/radix8.rs
```diff
@@ -0,0 +1,261 @@
...
+    input: &[In],
+    zero: Out,
+    params: &Radix8PNTTParams<C>,
+    mul_by_twiddle: M,
+) -> Vec<Out>
+where
+    C: Config,
+    In: Clone + Send + Sync,
+    Out: Clone + From<In> + CheckedAdd + CheckedMul + Send + Sync + Sum,
+    for<'a> &'a Out: From<&'a In>,
```
frozenspider: [Ditto](https://github.com/NethermindEth/zinc-plus/pull/51/changes#r2627024761), `In` and `Out` are the same thing

zip-plus/src/code/iprs.rs
```diff
@@ -0,0 +1,157 @@
...
+    fn new(poly_size: usize) -> Self {
+        assert_eq!(
+            poly_size % Config::N,
+            0,
+            "Polynomial size {} is not a multiple of row length {}",
+            poly_size,
+            Config::N
+        );
+
+        assert!(
```
frozenspider: Nit: `assert_eq!`

zip-plus/src/code/iprs.rs
```diff
@@ -0,0 +1,157 @@
...
+        );
+
+        Self {
+            pntt_params: Radix8PNTTParams::new(),
+            _phantom: Default::default(),
+        }
+    }
+
+    fn encode(&self, row: &[Zt::Eval]) -> Vec<Zt::Cw> {
+        assert!(
```
frozenspider: Ditto

zip-plus/src/code/iprs/pntt/radix8.rs
```diff
@@ -0,0 +1,261 @@
...
+        let sub_chunk_length = C::BASE_DIM * (1 << (3 * k));
+
+        // Work separately on combining each chunk of the next layer.
+        cfg_chunks_mut!(out, 8 * sub_chunk_length).for_each(|chunk: &mut [Out]| {
+            for i in 0..sub_chunk_length {
+                // Prepare subresults. Multiply them by the right roots of unity.
+                let subresults: [Out; 8] = array::from_fn(|j| {
+                    mul_by_twiddle.mul_by_twiddle(
+                        &chunk[j * sub_chunk_length + i],
+                        &params.roots_of_unity[j * i * (1 << (3 * (C::DEPTH - 1 - k)))],
```
frozenspider: I think it makes sense to precompute `1 << (3 * (C::DEPTH - 1 - k))` at `k` level, if not for performance, then for simplicity

zip-plus/src/code/iprs/pntt/radix8.rs
```diff
@@ -0,0 +1,261 @@
...
+fn base_multiply_into_output<In, Out, C, M>(
+    input: &[In],
+    params: &Radix8PNTTParams<C>,
+    zero: Out,
+    mul_by_twiddle: M,
+) -> Vec<Out>
+where
+    C: Config,
+    In: Clone + Send + Sync,
+    Out: Clone + From<In> + CheckedAdd + CheckedMul + Sum + Send + Sync + CheckedMul + Sum + Send,
```
frozenspider: A lot of duplicate types here.
To avoid this I always try to keep types order consistent 🙂 (e.g. `clone, send, sync` going last, also because they're least interesting)

Also, neither `CheckedMul` nor `Sum` are used.

zip-plus/src/code/iprs/pntt/radix8.rs
```diff
@@ -0,0 +1,261 @@
...
+    for<'a> &'a Out: From<&'a In>,
+    M: MulByTwiddle<Out, C::Int>,
+{
+    let mut output = vec![zero; C::M];
```
frozenspider: Instead of preallocating a mutable vector, we can do `cfg_into_iter!(0..C::M).map(|i| { ... }).collect()`

frozenspider (changes requested):

zip-plus/src/code/iprs/pntt/radix8/butterfly.rs
ElijahVlasov: I got rid of it completely in https://github.com/NethermindEth/zinc-plus/pull/53. So I suggest we don't touch it in this PR.

zip-plus/src/code/iprs/pntt/radix8/params.rs
ElijahVlasov: Is there a chance to make rustfmt/clippy catch that?

zip-plus/src/code/iprs/pntt/radix8/butterfly.rs
ElijahVlasov: Actually let's make the git history slightly nicer

zip-plus/src/code/iprs/pntt/radix8/params.rs
ElijahVlasov: Accessing a row of a `DenseRowMatrix` is clumsy now as it doesn't implement `Index` trait. Don't feel like using the `.as_rows()` iterator for randomly accessing a row. I'm putting a todo, once we have everything from `DenseRowMatrix` side we can switch.

zip-plus/src/code/iprs/pntt/radix8/butterfly.rs
```diff
@@ -0,0 +1,39 @@
...
+            *y = xs[1..]
+                .iter()
+                .zip(&butterfly[1..])
+                .fold(xs[0].clone(), |a, (x, &twiddle_idx)| {
+                    add!(a, &mul_by_twiddle.mul_by_twiddle(x, &twiddles[twiddle_idx]))
+                })
+        });
+}
+
+const BUTTERFLY_TABLE: [[usize; 8]; 8] = [
```
osdnk: Maybe instead of writing the whole matrix and then ignoring the first column, we can just write without the first column?

zip-plus/src/code/iprs/pntt/radix8/butterfly.rs
```diff
@@ -0,0 +1,39 @@
...
+    R: Clone + CheckedAdd,
+    M: MulByTwiddle<R, Twiddle>,
+{
+    ys.into_iter()
+        .zip(BUTTERFLY_TABLE.iter())
+        .for_each(|(y, butterfly)| {
+            *y = xs[1..]
+                .iter()
+                .zip(&butterfly[1..])
+                .fold(xs[0].clone(), |a, (x, &twiddle_idx)| {
```
osdnk: I understand that this is some micro-optimisation so that `butterfly` is iterated from `1` index. It doesn't seem to bring much, as we need to do cloning anyway. Maybe it's worth a comment? I find it relatively confusing.

zip-plus/src/code/iprs/pntt/radix8/butterfly.rs
```diff
@@ -0,0 +1,39 @@
+use num_traits::CheckedAdd;
+use zinc_utils::add;
+
+use super::MulByTwiddle;
+
+/// Apply butterfly given by `twiddles` to a slice
+/// of subresults in `x`. Use `mul_by_twiddle` as a means
+/// to multiply `Out` by `Twiddle`.
+pub(crate) fn apply_radix_8_butterflies<R, Twiddle, M>(
+    xs: &[R],
```
osdnk: Maybe I'm a bit biased towards some patterns I know from C (i.e. maybe it's different in Rust, so feel free to ignore my comment), but if we use `out-parameter` pattern (the result is given out as a mutable parameter), then the result (`ys`?) should go first.

zip-plus/src/code/iprs/pntt/radix8/butterfly.rs
```diff
@@ -0,0 +1,39 @@
...
+    ys: [&mut R; 8],
+    twiddles: &[Twiddle],
+    mul_by_twiddle: M,
```
osdnk: I think this M can be a ref (I tried changing everywhere and it seems fine)? Yet, please see also my other comment in `mul_by_twiddle.rs`

zip-plus/src/code/iprs/pntt/radix8/mul_by_twiddle.rs
```diff
@@ -0,0 +1,56 @@
...
+impl<F, Twiddle, T> MulByTwiddle<F, Twiddle> for FieldMulByTwiddle<F, T>
+where
+    F: PrimeField + FromWithConfig<T>,
+    Twiddle: Clone + Into<T>,
+    T: Clone + Send + Sync,
+{
+    fn mul_by_twiddle(&self, lhs: &F, twiddle: &Twiddle) -> F {
+        F::from_with_cfg(twiddle.clone().into(), &self.config) * lhs
+    }
+}
```
osdnk, line 57: I think I am very confused by this file. This is my understanding of this logic: 

Trait `MulByTwiddle` is a trait that is used for structs that serve as helpers for multiplication of Lhs and Twiddle. 

Trait is implemented for `FieldMulByTwiddle` and `MBSMulByTwiddle`. The latter also comes with a derived default implementation. 

There's also a phantom type that purpose here I cannot process yet.

With this understanding in mind, I wonder what the need is to define structs like `FieldMulByTwiddle` and `MBSMulByTwiddle`, as essentially they are static objects used for multiplying two elements. 

As an alternative, I think `MulByTwiddle` could be a trait of a similar nature as `MulByScalar`, i.e. implemented for `Lhs`. In this context, that would mean forwarding the logic from the `MulByScalar` trait + implementing a similar one for `Field` using `FromWithConfig`. 

The benefit (except for simplicity) would be that this struct wouldn't have to be passed on and through NTT execution (argument `M`).

zip-plus/src/code/iprs/pntt/radix8.rs
```diff
@@ -0,0 +1,262 @@
...
+                #[allow(unused_mut)] // false alarm
+                let ys: [&mut R; 8] = chunk
+                    .chunks_mut(sub_chunk_length)
+                    .map(|mut subchunk| &mut subchunk[i])
+                    .collect_vec()
+                    .try_into()
+                    .expect("We are guaranteed to have the right length here");
+
+                // Perform butterflies.
+                apply_radix_8_butterflies(&subresults, ys, &(C::TWIDDLES), mul_by_twiddle.clone());
```
osdnk: Here I am a bit confused again. I think this is the most important comment in this review.

It seems that now the pntt is performed in two stages, which are:
(i) we perform the first `mul_by_twiddle` with some roots of unity (mod p, lifted to ZZ),
(ii) we perform the second `mul_by_twiddle` for butterfly with twiddles (lifted to ZZ). 

As such, the inf norm of our code worsens/grows twice for each step (a bit more for (ii) due to 8 additions, but roughly similarly). 

According to my understanding of IPRS, this is unnecessary as those two steps can be combined (as they were in my implementation). Then, the norm will worsen two times less (in the bit length, as the lift to ZZ should happen after multiplying (mod p) twiddles with roots of unity).

Also, we can precompute twiddles for each layer with already multiplied by roots of unity (with some memory cost, but it's fine since we already store a lot of roots of unity), essentially eliminating the need for L79 - L93.

zip-plus/src/code/iprs/pntt/radix8.rs
```diff
@@ -0,0 +1,262 @@
...
+) -> Vec<R>
+where
+    C: Config,
+    R: Clone + CheckedAdd + CheckedMul + Sum + Send + Sync,
+    M: MulByTwiddle<R, C::Int>,
+{
+    cfg_into_iter!(0..C::OUTPUT_LEN)
+        .map(|i| {
+            let chunk = i / C::BASE_DIM;
+            let row = i % C::BASE_DIM;
```
osdnk: This is small, but can potentially save us a little bit `%` and `/` are done via so-called long division, which is especially over u128 can be done in hundreds of CPU cycles (it's still in ns, so not super important). Since `C::BASE_DIM` (and every dimension here) is a power of two, I think it's a reasonable practice to use `>>` and `<<` (1 CPU cycle) and store the size as a log. 

Typically, a compiler is smart enough to do it itself, but here it has no way to know that `C::BASE_DIM` is a power of two.

Same holds everywhere else.

zip-plus/src/code/iprs/pntt/radix8.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    #[test]
+    fn compare_to_arkworks_ntt_base_layer_multiply() {
+        compare_to_arkworks_ntt_base_layer_generic::<PnttConfigF2_16_1<1>>();
+        compare_to_arkworks_ntt_base_layer_generic::<PnttConfigF2_16_1<2>>();
+        compare_to_arkworks_ntt_base_layer_generic::<PnttConfigF2_16_1<3>>();
+    }
+
+    fn pntt_against_arkworks_generic<C: Config>()
```
osdnk, line 202: Great test, maybe some comment would be nice, i.e. why it should work and why it is important.

zip-plus/benches/zip_plus_benches.rs
```diff
@@ -56,26 +83,40 @@ impl RaaConfig for BenchRaaConfig {
...
     >(&mut group);
 
     group.finish();
 }
 
-criterion_group!(benches, zip_plus_benchmarks);
+fn zip_plus_benchmarks_iprs(c: &mut Criterion) {
+    let mut group = c.benchmark_group("Zip+ IPRS");
+
+    encode_rows::<BenchZipPlusTypes<i128, 32>, SomeIprsCode<1, _>, 16>(&mut group);
```
osdnk: I don't understand the value of this bench. IIUC, `encode_rows` chunks `2^16` evals into `2^8` chunks of len `2^8 = 2^{3 * 1 +5}` and sequentially encodes each row. There's no reason why this shouldn't be just `2^8` times slower than the encoding of a single row?

zip-plus/benches/zip_plus_benches.rs
```diff
@@ -56,26 +83,40 @@ impl RaaConfig for BenchRaaConfig {
...
+
+    encode_rows::<BenchZipPlusTypes<i128, 32>, SomeIprsCode<2, _>, 11>(&mut group);
+
+    encode_rows::<BenchZipPlusTypes<i128, 32>, SomeIprsCode<3, _>, 14>(&mut group);
```
osdnk, line 117: I think the intention was to encode a single row, so 11->10 and 14->13 I guess?

zip-plus/src/code/iprs/pntt/radix8/butterfly.rs
osdnk: Some ideas for optimisation would be to store inner-product early for y = 0 and y = 3, and then scale up after.

zip-plus/src/code/iprs/pntt/radix8/octet_reversal.rs
```diff
@@ -0,0 +1,64 @@
...
+#[allow(clippy::arithmetic_side_effects)]
+pub(crate) fn octet_reversal(x: usize, num_octets: usize) -> usize {
+    assert_ne!(num_octets, 0, "The number of octets cannot be zero");
+
+    let mut result = 0;
+
+    let mut left_octet_mask = 0b111;
+    let mut right_octet_mask = 0b111 << (3 * (num_octets - 1));
+
+    for i in 0..(num_octets / 2 + num_octets % 2) {
```
osdnk, line 14:
```suggestion
    for i in 0..num_octets >> 1 {
```

If you care about 🏎️  speed 🏎️  :P 

The inner (if num_octets is odd) octet doesn't have to be swapped with itself :)

zip-plus/src/code/iprs/pntt/radix8/params.rs
```diff
@@ -0,0 +1,157 @@
...
+    const TWIDDLES: [Self::Int; 8];
+
+    /// The length of the pseudo NTT's input.
+    const INPUT_LEN: usize = Self::BASE_LEN * (1 << (3 * Self::DEPTH));
+    /// The length of the pseudo NTT's output.
+    const OUTPUT_LEN: usize = Self::BASE_DIM * (1 << (3 * Self::DEPTH));
+
+    /// A helper to get an integer representation that
+    /// lies in the range `[-(p - 1)/2, (p - 1)/2]` from a field element.
+    fn field_to_int_normalized(x: Self::Field) -> Self::Int;
```
osdnk: This seems to me like a trait candidate? 

```rust
pub trait ToIntNormalized<IntOutput> {
   fn to_int_normalized(x: self) ->  IntOutput;
}
```

zip-plus/src/code/iprs.rs
```diff
@@ -0,0 +1,149 @@
...
+use std::{iter::Sum, marker::PhantomData, ops::AddAssign};
+use zinc_utils::{from_ref::FromRef, mul_by_scalar::MulByScalar};
+
+pub use pntt::radix8::params::{PnttConfigF2_16_1, Radix8PnttParams};
+
+/// Pseudo Reed-Solomon encoder over the integers. Internally uses a
+/// radix-8 NTT-style recursion with a base Vandermonde matrix sized
+/// `base_len x base_dim` (defaults to 64x32).
+#[derive(Debug, Clone)]
+pub struct IprsCode<Zt: ZipTypes, Config: pntt::radix8::params::Config> {
```
osdnk:
```suggestion
pub struct IprsCode<Zt: ZipTypes, Config: Config> {
```

maybe rename to `PnttConfig`?

osdnk (changes requested):
Great work! Here is a bunch of comments. I can adopt them if we're on the same page.

The most important comments are in files radix8.rs and mul_by_twiddle.rs.

zip-plus/src/code/iprs/pntt/radix8.rs
```diff
@@ -0,0 +1,262 @@
...
+            let chunk = i / C::BASE_DIM;
+            let row = i % C::BASE_DIM;
+
+            // If we'd done all the divide steps of the NTT recursively
+            // we'd end up with chunks of original indices
+            // combined together according to their `3 * C::DEPTH`
+            // least significant bits. Moreover, the value of these
+            // least significant bits correspond to the number of the chunk
+            // in octet-reverse order.
+            let oct_rev_chunk = octet_reversal(chunk, C::DEPTH);
```
osdnk, line 123: Not an issue, but rather an idea/suggestion: 
What is wrong with just letting it stay in "octet-reversed-order"? I think this representation might be even better for further testing/evaluating for cache locality?

zip-plus/src/code/iprs/pntt/radix8/params.rs
frozenspider: I don't think there is, sadly

zip-plus/src/code/iprs/pntt/radix8/octet_reversal.rs
frozenspider: Good point!
Note though that `num_octets >> 1` is a terrible code compared to `num_octets / 2`, and they both compile [exactly the same](https://godbolt.org/z/rzvMKqhx8)

zip-plus/src/code/iprs/pntt/radix8/butterfly.rs
frozenspider: Yeah. There's no well established fixed order of variables in Rust as there is in C, but mixing input and output variables is not a good idea.

zip-plus/src/code/iprs/pntt/radix8/mul_by_twiddle.rs
frozenspider: You're right, even `MulByScalar` alone would be enough to cover this case. However, due to Rust constraints, we can't really define `impl<F: FromWithConfig<i64>> MulByScalar<&i64> for F` as this would be conflicting with e.g. `impl MulByScalar<&i64> for i128` - Rust fears an upstream crate might someday implement `impl FromWithConfig<i64> for i128` and that would break things. Same thing would apply if we use `MulByTwiddle` in this fashion.
I'll think of how else this could be structured, but for now I would go with the current implementation.

zip-plus/src/code/iprs/pntt/radix8/octet_reversal.rs
frozenspider: Turns out, we do need `+ num_octets % 2`, otherwise the mask bit shift doesn't do the right thing

zip-plus/src/code/iprs/pntt/radix8/params.rs
frozenspider: We could do it as a separate trait, yeah, but I don't really see much value in doing so. There's just one implementation, it's not used outside of IPRS, and extracting it into a trait would just complicate type declarations.

zip-plus/src/code/iprs/pntt/radix8.rs
frozenspider: Can't really say anything about math implications of this. If you provide a diff of how the multiplication by twiddle would look like in this case (I assume `oct_rev_chunk | ((col + 1) << (3 * C::DEPTH))` would no longer be right?), I'd be happy to integrate it.

zip-plus/benches/zip_plus_benches.rs
frozenspider: It doesn't hurt to have a benchmark that measures this. If this ever becomes superlinear due to a mistake on our part, we'll detect it quickly.

zip-plus/src/code/iprs/pntt/radix8.rs
frozenspider: Again, my understanding of IPRS is not good enough to find a middle ground between current implementation and your old one. Would you mind helping me out? Thanks!

zip-plus/benches/zip_plus_benches.rs
frozenspider: As of now, this causes an overflow. Could this be because of the doubled norm growth that you mentioned?

frozenspider:
Addressed most of the comments.
Also, `Config::Int` type was both `From<i64>` and `Into<i64>` and that almost always limited us to just `i64` that it actually was. As such, replaced it with a type alias `pub type PnttInt = i64`

zip-plus/src/code/iprs/pntt/radix8/params.rs
osdnk: For me, this is more of an abstraction issue. "Ability to normalise" seems to me more like a property of a field, not the IPRS code. But this is only my perspective -- I don't care that much ^^

zip-plus/src/code/iprs/pntt/radix8.rs
osdnk: Let's not worry about this rn

zip-plus/benches/zip_plus_benches.rs
osdnk: I think so! My calculation shows that it shouldn't overflow here (assuming single growth)

zip-plus/src/code/iprs/pntt/radix8/mul_by_twiddle.rs
osdnk: Oh, interesting. Thanks for that.

zip-plus/src/code/iprs/pntt/radix8.rs
osdnk: I can, of course. Do you want me to do the coding here? If yes, I can try tomorrow, but more likely on Monday or even Tuesday (we have Christmas now).

frozenspider: Whatever way you prefer works by me. By all means, enjoy your holiday!

zip-plus/src/code/iprs/pntt/radix8/params.rs
frozenspider: Yeah, I get what you mean. Given the way it's implemented right now though, making it truly generic over any field would be way too much effort, so I propose to postpone it until we need it somewhere else. Leaving a `todo` would be reasonable though.

zip-plus/src/code/iprs/pntt/radix8/octet_reversal.rs
osdnk: ¯\_(ツ)_/¯ interesting

osdnk (approved):
I think I am happy with the changes and answers according to my previous review.

Some points for further optimisations are:

- `BUTTERFLY_TABLE` has a lot of inside-row repetitions we can take advantage of
- The dimensions of the default base case are relatively large (32x64), and it seems that now the base case takes the majority of the time. Maybe it's worth rethinking?

zip-plus/benches/zip_plus_benches.rs
frozenspider: Note that it still overflows even with double norm growth addressed

frozenspider (approved):

PR #52, opened 2025-12-15
IPRS optimisations

ElijahVlasov:
1. Introduce means to specify a better mul-by-scalar algorithm for binary polynomials (similar to what we did with inner product).
2. Remove unneeded zero argument to all of the pntt subprocedures.
3. Do a bench for i64 coefficients of codewords rather than i128

### DIFF
--- a/poly/src/univariate/dense.rs
+++ b/poly/src/univariate/dense.rs
@@ -1,7 +1,7 @@
 use crate::{ConstCoeffBitWidth, EvaluatablePolynomial, EvaluationError, Polynomial};
 use core::slice;
 use crypto_primitives::{
-    FixedSemiring, FromWithConfig, IntoWithConfig, PrimeField, Ring, Semiring,
+    FixedSemiring, FromWithConfig, IntoWithConfig, PrimeField, Ring, Semiring, boolean::Boolean,
 };
 use itertools::Itertools;
 use num_traits::{CheckedAdd, CheckedMul, CheckedNeg, CheckedSub, One, Zero};
@@ -18,7 +18,7 @@ use zinc_transcript::traits::ConstTranscribable;
 use zinc_utils::{
     from_ref::FromRef,
     inner_product::{InnerProduct, InnerProductError},
-    mul_by_scalar::MulByScalar,
+    mul_by_scalar::{MulByScalar, WideningMulByScalar},
     named::Named,
     projection_to_field::ProjectionToField,
 };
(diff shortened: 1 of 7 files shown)


frozenspider:
Updated this PR according to the latest state of IPRS codes

poly/src/univariate/dense.rs
```diff
@@ -627,3 +627,31 @@ where
...
+impl<Rhs, Output, const DEGREE_PLUS_ONE: usize>
+    WideningMulByScalar<DensePolynomial<Boolean, DEGREE_PLUS_ONE>, Rhs>
+    for BinaryPolyWideningMulByScalar<Output>
+where
+    Rhs: Copy,
+    Output: From<Rhs> + Send + Sync + Default + Copy + Zero,
+{
+    type Output = DensePolynomial<Output, DEGREE_PLUS_ONE>;
+
+    fn mul_by_scalar_widen(
```
osdnk, line 643: Not a blocker here, but again, as in the previous review, this pattern smells wrong to me.

zip-plus/src/code/iprs/pntt/radix8/mul_by_twiddle.rs
```diff
@@ -51,7 +55,25 @@ where
...
         F::from_with_cfg(twiddle.clone().into(), &self.config) * lhs
     }
 }
+
+#[derive(Clone, Default, Copy)]
+pub struct WideningMulByTwiddle<WM>(PhantomData<WM>);
+
+impl<Lhs, Twiddle, WM> MulByTwiddle<Lhs, Twiddle> for WideningMulByTwiddle<WM>
+where
+    WM: WideningMulByScalar<Lhs, Twiddle>,
```
osdnk: can we avoid using those abbreviations? Where `M` itself was somewhat clear, I think `WM`, `MI`, `MO` are a bit too much to my taste.

zip-plus/src/code/iprs/pntt/radix8.rs
```diff
@@ -134,9 +144,17 @@ where
...
                 },
-            )
-        })
-        .collect()
+            );
+
+            *out = MaybeUninit::new(result);
+        });
+
+    // Safety: We initialized all elements in the output.
```
osdnk: is it really giving us anything?

osdnk (approved):
No serious worries from my end, the code seems right. Only some minor aesthetic suggestions.

poly/src/univariate/dense.rs
ElijahVlasov: I've started addressing similar situations in this pr https://github.com/NethermindEth/zinc-plus/pull/54. I pinky promise to address this too.

PR #44, opened 2025-12-02
`InnerProduct` and `ProjectableToField` traits reworked

ElijahVlasov:
Currently, implementations of `InnerProduct` and `ProjectableToField` can't be specialised for certain classes of types if a general implentation provided. For instance, `DensePolynomial<Boolean, _>` can't have its own specialised inner product if an implementation is provided for `DensePolynomial<R, _> where R: Semiring`. 

Inner product and projection to field shouldn't be tied to a (generic) type, rather it should be possible to switch between different implementations at will when it's necessary. It does clutter some function calls but I believe it's worth having some freedom.

As a bonus: benchmarks of evaluation phase for zip+ are improved by 90%.

PR #54, opened 2025-12-18
Inner products and binary polynomials v2

ElijahVlasov:
This PR partly reverts back https://github.com/NethermindEth/zinc-plus/pull/44. 

The need to fill in inner products and projections at each call site is unpleasant and it's sole purpose was to differentiate `InnerProduct` and `ProjectableToField` for `DensePolynomial<Boolean, DEGREE_PLUS_ONE>`.  For that reason this PR introduces a newtype-style wrapper for `DensePolynomial<Boolean, DEGREE_PLUS_ONE>` called `BinaryPoly<DEGREE_PLUS_ONE>`:
```rust
#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
#[repr(transparent)]
pub struct BinaryPoly<const DEGREE_PLUS_ONE: usize>(DensePolynomial<Boolean, DEGREE_PLUS_ONE>);
```
It will allow to easily do things differently for binary polynomials while delegating some of the functionality to `DensePolynomial<Boolean, DEGREE_PLUS_ONE>`.

Also reworked `MulByTwiddle` to be fully static, without a need to pass a reference `&mul_by_twiddle` around.


poly/src/univariate/binary.rs
```diff
@@ -0,0 +1,507 @@
...
+}
+
+impl<const DEGREE_PLUS_ONE: usize> Display for BinaryPoly<DEGREE_PLUS_ONE> {
+    #[inline(always)]
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl<const DEGREE_PLUS_ONE: usize> Hash for BinaryPoly<DEGREE_PLUS_ONE> {
```
frozenspider: I believe this can be `#[derive]`'d

poly/src/univariate/binary.rs
```diff
@@ -0,0 +1,507 @@
...
+}
+
+impl<const DEGREE_PLUS_ONE: usize> One for BinaryPoly<DEGREE_PLUS_ONE> {
+    #[inline(always)]
+    fn one() -> Self {
+        Self(DensePolynomial::one())
+    }
+}
+
+impl<const DEGREE_PLUS_ONE: usize> Add for BinaryPoly<DEGREE_PLUS_ONE> {
```
frozenspider: We could significantly reduce the amount of boilerplate we need with the use of https://crates.io/crates/derive_more

utils/src/field/const_monty.rs
```diff
@@ -70,7 +85,8 @@ mod tests {
         // Create a sample field element and an Int value
         let sampled = F::from(5_u64);
 
-        let projection_fn = SimpleProjection::prepare_projection(&sampled);
+        let projection_fn =
+            <Int<{ U128::LIMBS }> as ProjectableToField<F>>::prepare_projection(&sampled);
```
frozenspider: Nit: We don't need to explicitly specify `ProjectableToField`
```suggestion
            Int::<{ U128::LIMBS }>::prepare_projection(&sampled);
```

poly/src/univariate/binary.rs
```diff
@@ -0,0 +1,507 @@
...
+                    } else {
+                        R::zero()
+                    },
+                    R::one(),
+                ),
+                |(mut acc, mut pow), coeff| {
+                    pow = mul!(pow, point);
+
+                    if coeff.inner() {
+                        acc = add!(acc, &pow);
```
frozenspider: I think we should return `EvaluationError::Overflow` here instead of panicking

utils/src/inner_product.rs
```diff
@@ -50,174 +64,273 @@ where
...
+macro_rules! impl_inner_product_for_primitives {
+    ($($t:ty),*) => {
+       $(
+           impl<T, Out> InnerProduct<T, Out> for $t
+           where
+               T: Zero + One + PartialEq,
+               Out: Zero + One + From<$t> + for<'a> From<&'a T> + CheckedMul,
+           {
+                fn inner_product(&self, point: &[T], _zero: Out) -> Result<Out, InnerProductError> {
+                        if point.len() != 1 {
```
frozenspider: Style: Double indentation, here and below.
Note: `rustfmt` doesn't really format macro content

utils/src/inner_product.rs
```diff
@@ -50,174 +64,273 @@ where
...
-            Boolean::from(true),
-        ];
-        let rhs = [1i128, 2, 3, 4];
+impl<T, Rhs, Out> InnerProduct<Rhs, Out> for ForceUncheckedInnerProduct<T>
+where
+    T: InnerProductUnchecked<Rhs, Out>,
+{
+    #[inline(always)]
+    fn inner_product(&self, rhs: &[Rhs], zero: Out) -> Result<Out, InnerProductError> {
+        self.0.inner_product_unchecked(rhs, zero)
```
frozenspider: Given that we still need this sort of wrappers, maybe it makes sense to keep an `InnerProduct` to be external static structs without `&self` (similar to what it was before)?

poly/src/univariate/binary.rs
```diff
@@ -0,0 +1,507 @@
...
+    add,
+    from_ref::FromRef,
+    inner_product::{InnerProduct, InnerProductError, InnerProductUnchecked},
+    mul,
+    mul_by_scalar::WideningMulByScalar,
+    named::Named,
+    projectable_to_field::ProjectableToField,
+};
+
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
```
frozenspider: I have my doubts on whether we really want this to be `Copy`, `DensePolynomial` could be expensive to copy around. If it's not though, we should declare `DensePolynomial: Copy` as well.
Could you bench both approaches?

zip-plus/src/code/iprs/pntt/radix8/mul_by_twiddle.rs
```diff
@@ -1,32 +1,70 @@
...
+    fn new_ref(value: &T) -> &Self {
+        // Safety: ForceWideningMulByTwiddle is #[repr(transparent)] and is
+        // guaranteed to have the same memory layout as T
+        unsafe { &*(value as *const T as *const Self) }
+    }
+}
+
+impl<T, WideningMBS> Mul<&PnttInt> for &ForceWideningMulByTwiddle<T, WideningMBS>
+where
+    WideningMBS: WideningMulByScalar<T, PnttInt>,
```
frozenspider: Ditto about static structs vs wrapping everything

frozenspider (changes requested):

poly/src/univariate/binary.rs
ElijahVlasov: Yeah, I thought about it. Was not sure what the policy is for adding new deps :upside_down_face:

frozenspider: To my knowledge, we don't have any specific limitations 🙂

poly/src/univariate/binary.rs
```diff
@@ -0,0 +1,507 @@
...
+        if DEGREE_PLUS_ONE.is_one() {
+            return Ok(R::zero());
+        }
+
+        let result = self.0.coeffs[1..]
+            .iter()
+            .fold(
+                (
+                    if self.0.coeffs[0].inner() {
+                        R::one()
```
frozenspider: Once https://github.com/NethermindEth/crypto-primitives/pull/21 merges in, this could be slightly more concise

poly/src/univariate/binary.rs
```diff
@@ -0,0 +1,397 @@
...
+    /// If the input has fewer than N+1 coefficients, the remaining slots will
+    /// be filled with zeros. If the input has more than N+1 coefficients,
+    /// it will panic.
+    #[inline(always)]
+    pub fn new(coeffs: impl AsRef<[Boolean]>) -> Self {
+        Self(DensePolynomial::new(coeffs))
+    }
+}
+
+impl<const DEGREE_PLUS_ONE: usize> BinaryPoly<DEGREE_PLUS_ONE> {
```
frozenspider: These two `impl` blocks should be one

poly/src/univariate/binary.rs
```diff
@@ -0,0 +1,397 @@
...
+    type Output = Self;
+
+    #[allow(clippy::arithmetic_side_effects)]
+    #[inline(always)]
+    fn sub(self, rhs: &'a Self) -> Self::Output {
+        Self(self.0 - rhs.0)
+    }
+}
+
+impl<const DEGREE_PLUS_ONE: usize> Mul<BinaryPoly<DEGREE_PLUS_ONE>>
```
frozenspider: We don't need to explicitly declare it as `Mul<BinaryPoly<DEGREE_PLUS_ONE>>`, definition is `Mul<Rhs = Self>` so just `Mul` will do

zip-plus/src/pcs/phase_verify.rs
```diff
@@ -148,7 +147,7 @@ impl<Zt: ZipTypes, Lc: LinearCode<Zt>> ZipPlus<Zt, Lc> {
                 .map(Zt::Comb::from_ref)
                 .map(|p| Zt::CombDotChal::inner_product(&p, alphas, Zt::CombR::ZERO))
                 .try_collect()?;
-            I::inner_product(&column_entries, coeffs, Zt::CombR::ZERO)?
+            MBSInnerProductChecked::inner_product(&column_entries, coeffs, Zt::CombR::ZERO)?
```
frozenspider: Hmm, does it make sense to hardcode `...Checked` here, as opposed to generic `I` like it was before?
I don't have a strong opinion here, leaving it up to your discretion.

frozenspider (approved):
Great job!

zip-plus/src/pcs/phase_verify.rs
ElijahVlasov: I've offloaded the choice of this Inner product to `ZipTypes`.

ElijahVlasov:
Re: `Copy` trait for `DensePolynomial`.

Removing it breaks something in `ZipTypes` so I left a todo regarding that.

PR #61, opened 2026-01-09
Sumcheck benchmark and some optimisations

ElijahVlasov:
This PR introduces a random field sumcheck protocol that using a random field configuration projects the input MLEs onto that field and performs the ordinary sumcheck over a field.

The keypoints:

1. `MultilinearExtensionWithConfig`: allows to store only Montgomery representations in an MLE structure.
2. `project_coeffs`: allows to project onto a field all the coefficients of a dense MLE.
3. `RFSumcheck`: the random field sumcheck protocol itself.


piop/src/sumcheck/prover.rs
```diff
@@ -5,22 +5,22 @@ use crypto_primitives::PrimeField;
...
 
 /// Sumcheck Prover State.
-pub struct ProverState<F> {
+pub struct ProverState<F: PrimeField> {
     /// Sampled randomness given by the verifier.
     pub randomness: Vec<F>,
     /// Stores the list of multilinear extensions
     /// the sumcheck polynomial is comprised of.
-    pub mles: Vec<DenseMultilinearExtension<F>>,
+    pub mles: Vec<DenseMultilinearExtension<F::Inner>>,
```
frozenspider, line 23: What's the rationale behind storing `DenseMultilinearExtension<F::Inner>` instead of `<F>`? Does it affect performance in any way?

poly/src/utils.rs
```diff
@@ -105,3 +106,107 @@ where
...
+        // for the current step we will need
+        // if x_0 = 0:   (1-r0) * [b_1, ..., b_k]
+        // if x_0 = 1:   r0 * [b_1, ..., b_k]
+        // let mut res = vec![];
+        // for &b_i in buf.iter() {
+        //     let tmp = r[0] * b_i;
+        //     res.push(b_i - tmp);
+        //     res.push(tmp);
+        // }
+        // *buf = res;
```
osdnk: I think this code should be removed.

piop/src/random_field_sumcheck.rs
```diff
@@ -0,0 +1,228 @@
...
+        num_vars: usize,
+        degree: usize,
+        claimed_sum: F,
+        proof: &RFSumcheckProof<F, R>,
+        field_cfg: F::Config,
+    ) -> Result<SubClaim<F>, RFSumcheckError<F>>
+    where
+        F::Inner: ConstTranscribable + ConstIntSemiring,
+    {
+        // Simulate getting the projecting element.
```
osdnk:
```suggestion
        // Simulate getting the projecting element
        // Verifier does not use that element as it verifies only over RC,
        // but we keep it here for stability of FS sampling. 
```

piop/src/sumcheck/prover.rs
osdnk: I think that this comment clarifies it 
https://github.com/NethermindEth/zinc-plus/pull/61/changes#diff-4a2886a40f9865228a50b27fb5a02c3a0293d270671fade19d5f529c46eedb97R49

frozenspider: Not really, it describes how we're using `F::Inner` + `F::Config`, but not why are we doing it.
I've run some benches and it seems that cloning `F::Config` when constructing a vector really does increase sumcheck prover time by ~20% 😞 
We could make another primitive to work around that, but for now let's go with the current approach - just please add comments as to why we're doing it this way. Also, I believe there are places where `F::Inner` optimization is superfluous, but let me comment separately on that.

ElijahVlasov: Another point which I don't have evidence for at hand is that it allows to save some memory as we don't need to store a field config with each element in the vector of MLE evaluations. Unless the compiler is smart enough (which I don't think it is) and it figures out that all the configs are the same. And, e.g., MontyParams are quite heavy

piop/src/random_field_sumcheck.rs
```diff
@@ -0,0 +1,228 @@
...
+
+impl<F: PrimeField, R> RFProverState<F, R> {
+    pub fn new(sumcheck_prover_state: ProverState<F>) -> Self {
+        Self(sumcheck_prover_state, Default::default())
+    }
+}
+
+impl<F: FromPrimitiveWithConfig, R: Semiring + ProjectableToField<F>> RFSumcheck<F, R> {
+    /// Random field sumcheck prover.
+    /// Samples a random field element, projects the input mles
```
frozenspider: Nit: MLEs

poly/src/utils.rs
```diff
@@ -105,3 +106,107 @@ where
...
+}
+
+/// This function build the eq(x, r) polynomial for any given r, and output the
+/// evaluation of eq(x, r) in its vector form.
+///
+/// Evaluate
+///      eq(x,y) = \prod_i=1^num_var (x_i * y_i + (1-x_i)*(1-y_i))
+/// over r, which is
+///      eq(x,y) = \prod_i=1^num_var (x_i * r_i + (1-x_i)*(1-r_i))
+pub fn build_eq_x_r_inner_vec<F>(r: &[F], cfg: &F::Config) -> Result<Vec<F::Inner>, ArithErrors>
```
frozenspider: Nit: Looks like this doesn't need to be public

piop/src/random_field_sumcheck.rs
```diff
@@ -0,0 +1,228 @@
...
+    ///   comes in this argument.
+    pub fn prove_as_subprotocol(
+        transcript: &mut impl Transcript,
+        mles: Vec<DenseMultilinearExtension<R>>,
+        mles_f: Vec<DenseMultilinearExtension<F::Inner>>,
+        nvars: usize,
+        degree: usize,
+        comb_fn: impl Fn(&F, &[F]) -> F + Send + Sync,
+        field_cfg: F::Config,
+    ) -> (RFSumcheckProof<F, R>, RFProverState<F, R>)
```
frozenspider, line 74: (Applies to both this and `MLSumcheck`)
It seems that the prover state returned from `prove_as_subprotocol` is always discarded. Maybe not return it then?

frozenspider (approved):

piop/src/random_field_sumcheck.rs
ElijahVlasov: It starts to be really important once we compose the sumcheck with the rest of the protocol (which we haven't done yet). As we need to continue to deal with the MLEs and with the verifier's randomness produced during sumcheck (aka evaluation point) to then proof that mles evaluated correctly using the PCS.

frozenspider (approved):
Awesome!

PR #64, opened 2026-01-20
Do not send constant term in sumchech

osdnk:
- Prover messages now omit the constant term entirely; the prover records the asserted sum from the first round and attaches it to `SumcheckProof` so `extract_sum` is just a lookup.
- Verifier stores truncated round polynomials, reconstructs the missing constant assuming P(0)+P(1)=expected, and evaluates the rebuilt NatEvaluatedPoly; transcript absorption updated to match the shorter payload.
- Protocol structures adjusted: `ProverMsg` now wraps a `Vec<F>`, `SumcheckProof` carries `(rounds, asserted_sum)`, and zero-variable/construction sites updated accordingly.
- Tests rewritten to reflect the new semantics (tampering/incorrect sums lead to different subclaims rather than immediate rejection) plus a guard that the constant term is absent from prover messages.

MacBook with M1 Pro
image


ElijahVlasov:
Could you pls resolve the conflicts first?

piop/src/sumcheck.rs
```diff
@@ -25,15 +25,12 @@ pub struct MLSumcheck<F>(PhantomData<F>);
 
 /// Proof generated by the sumcheck prover.
 #[derive(Clone, Debug, PartialEq)]
-pub struct SumcheckProof<F>(pub Vec<ProverMsg<F>>);
+pub struct SumcheckProof<F>(pub Vec<ProverMsg<F>>, pub F);
```
ElijahVlasov: Since it now has more than one fields can we have names for them?

piop/src/sumcheck.rs
```diff
@@ -25,15 +25,12 @@ pub struct MLSumcheck<F>(PhantomData<F>);
...
-where
-    F: for<'a> Add<&'a F, Output = F> + Clone,
-{
+impl<F: Clone> SumcheckProof<F> {
     /// Derive the claimed sum from the sumcheck protocol.
     pub fn extract_sum(&self) -> F {
-        self.0[0].0.evaluations[0].clone() + &self.0[0].0.evaluations[1]
+        self.1.clone()
     }
 }
```
ElijahVlasov: Probably it isn't needed anymore as we have the sum explicitly in the proof

ElijahVlasov (changes requested):
LGTM! Let's fix the tiny nits and make clippy happy

piop/src/sumcheck.rs
osdnk: yup

piop/src/sumcheck.rs
osdnk: yup

ElijahVlasov (approved):

PR #67, opened 2026-01-23
Comb features a bit

ElijahVlasov:


transcript/Cargo.toml
```diff
@@ -23,5 +23,5 @@ num-traits = { workspace = true }
 workspace = true
 
 [features]
-asm = ["sha3/asm"]
+sha3-asm = ["sha3/asm"]
```
frozenspider, line 26: Doesn't it make more sense to have a single umbrella `asm` feature that governs all asm-related things? That's what most published crates do

ElijahVlasov: Maybe. The only asm thing we have is sha3-asm so I wondered if it would be better to emphasise that. We can always roll back to "asm" if we get new asm things

PR #68, opened 2026-01-23
SIMD-accelerated binary polynomials

osdnk:
This branch adds SIMD optimisations for binary polynomial operations, with performance improvements for polynomial widening.

New SIMD implementation stores binary polynomials as compact u64 bitmasks instead of Boolean arrays
Architecture-specific optimisations for ARM NEON and x86 AVX-512. 

## Feature flag (`simd`)
Default: uses original `BinaryRefPoly` (renamed from `BinaryPoly`)
With `simd`: swaps in `BinaryU64Poly`
API stays identical, so existing code works unchanged

## Implementation details
ARM: Uses SIMD widening ops for most sizes, 16KB lookup table for N=64
x86: 4-way unrolled loop with native mask registers
Both use branchless selection for maximum throughput

NEON 
Pasted Graphic

AVX512
Pasted Graphic 1


poly/src/univariate/binary_u64.rs
```diff
@@ -0,0 +1,744 @@
...
+        // addition in GF(2) is XOR
+        Self(self.0 ^ rhs.0)
+    }
+}
+
+impl<'a, const DEGREE_PLUS_ONE: usize> Sub<&'a Self> for BinaryU64Poly<DEGREE_PLUS_ONE> {
+    type Output = Self;
+
+    #[allow(clippy::arithmetic_side_effects, clippy::suspicious_arithmetic_impl)]
+    #[allow(clippy::arithmetic_side_effects, clippy::suspicious_arithmetic_impl)]
```
frozenspider, line 136: Double annotation

poly/src/univariate/binary_u64.rs
```diff
@@ -0,0 +1,744 @@
...
+    }
+}
+
+impl<const DEGREE_PLUS_ONE: usize> Mul for BinaryU64Poly<DEGREE_PLUS_ONE> {
+    type Output = Self;
+
+    #[allow(clippy::arithmetic_side_effects)]
+    #[inline(always)]
+    fn mul(self, _rhs: Self) -> Self::Output {
+        unimplemented!("Multiplication for BinaryU64Poly is not implemented yet");
```
frozenspider: Nit: Delegate to `Mul<&'a Self>` (or vice versa, since it's a `Copy` primitive)

poly/src/univariate/binary_u64.rs
```diff
@@ -0,0 +1,744 @@
...
+use zinc_utils::{
+    from_ref::FromRef,
+    inner_product::{BooleanInnerProductUncheckedAdd, InnerProduct, InnerProductError},
+    mul_by_scalar::WideningMulByScalar,
+    named::Named,
+    projectable_to_field::ProjectableToField,
+};
+
+#[derive(
+    Add,
```
frozenspider: Why are we deriving so much stuff if we implement them directly? I assume deriving here does nothing?

poly/src/univariate/binary_u64.rs
```diff
@@ -0,0 +1,744 @@
...
+    fn mul_assign(&mut self, _rhs: Self) {
+        unimplemented!("Multiplication for BinaryU64Poly is not implemented yet");
+    }
+}
+
+impl<'a, const DEGREE_PLUS_ONE: usize> MulAssign<&'a Self> for BinaryU64Poly<DEGREE_PLUS_ONE> {
+    #[inline(always)]
+    fn mul_assign(&mut self, _rhs: &'a Self) {
+        unimplemented!("Multiplication for BinaryU64Poly is not implemented yet");
+        // self.0.mul_assign(&rhs.0);
```
frozenspider, line 217: Code leftover

poly/src/univariate/binary_u64.rs
```diff
@@ -0,0 +1,744 @@
...
+    fn sub_assign(&mut self, rhs: &'a Self) {
+        // subtraction in GF(2) is XOR
+        self.0 ^= rhs.0;
+    }
+}
+
+impl<const DEGREE_PLUS_ONE: usize> MulAssign for BinaryU64Poly<DEGREE_PLUS_ONE> {
+    #[inline(always)]
+    fn mul_assign(&mut self, _rhs: Self) {
+        unimplemented!("Multiplication for BinaryU64Poly is not implemented yet");
```
frozenspider, line 209: Ditto about delegation

poly/src/univariate/binary_u64.rs
```diff
@@ -0,0 +1,744 @@
...
+        // self.0.mul_assign(&rhs.0);
+    }
+}
+
+impl<const DEGREE_PLUS_ONE: usize> CheckedAdd for BinaryU64Poly<DEGREE_PLUS_ONE> {
+    #[inline(always)]
+    fn checked_add(&self, other: &Self) -> Option<Self> {
+        // addition in GF(2) is XOR
+        Some(Self(self.0 ^ other.0))
```
frozenspider: Checked addition is not XOR, it should return `None` if 1 + 1 bits are present - we're not working in GF(2) after all

poly/src/univariate/binary_u64.rs
```diff
@@ -0,0 +1,744 @@
...
+    }
+}
+
+impl<const DEGREE_PLUS_ONE: usize> CheckedSub for BinaryU64Poly<DEGREE_PLUS_ONE> {
+    #[inline(always)]
+    fn checked_sub(&self, other: &Self) -> Option<Self> {
+        // subtraction in GF(2) is XOR
+        Some(Self(self.0 ^ other.0))
```
frozenspider: Ditto, `0 - 1 = None`

poly/src/univariate/binary_u64.rs
```diff
@@ -0,0 +1,744 @@
...
+    #[inline(always)]
+    fn sum<I: Iterator<Item = &'a Self>>(iter: I) -> Self {
+        Self(iter.map(|x| &x.0).sum())
+    }
+}
+
+impl<'a, const DEGREE_PLUS_ONE: usize> Product<&'a Self> for BinaryU64Poly<DEGREE_PLUS_ONE> {
+    #[inline(always)]
+    fn product<I: Iterator<Item = &'a Self>>(iter: I) -> Self {
+        Self(iter.map(|x| &x.0).product())
```
frozenspider: Why are sum and product implemented as sum/product of underlying integers?

poly/src/univariate/binary_u64.rs
```diff
@@ -0,0 +1,744 @@
...
+
+impl<const DEGREE_PLUS_ONE: usize> Semiring for BinaryU64Poly<DEGREE_PLUS_ONE> {}
+
+impl<const DEGREE_PLUS_ONE: usize> Distribution<BinaryU64Poly<DEGREE_PLUS_ONE>>
+    for StandardUniform
+{
+    #[inline(always)]
+    fn sample<Gen: Rng + ?Sized>(&self, rng: &mut Gen) -> BinaryU64Poly<DEGREE_PLUS_ONE> {
+        let coeffs: [Boolean; DEGREE_PLUS_ONE] = rng.random();
+        BinaryU64Poly::new(coeffs)
```
frozenspider: Wouldn't it be better to sample u64 and zero out leading bits?
(I think this is only used in tests though, so it's a nitpick)

poly/src/univariate/binary_u64.rs
```diff
@@ -0,0 +1,744 @@
...
+    #[inline(always)]
+    fn inner_product(
+        lhs: &BinaryU64Poly<DEGREE_PLUS_ONE>,
+        rhs: &[Rhs],
+        zero: Out,
+    ) -> Result<Out, InnerProductError> {
+        let lhs = DensePolynomial::<Boolean, DEGREE_PLUS_ONE>::new(array::from_fn(|i| {
+            Boolean::new((lhs.0 & (1 << i)) != 0)
+        })
+            as [Boolean; DEGREE_PLUS_ONE]); // idk
```
frozenspider: Either `array::from_fn::<_, DEGREE_PLUS_ONE, _>` or `as [_; DEGREE_PLUS_ONE]` (I prefer the former as it can't be confused with type casting)

poly/src/univariate/binary_u64.rs
```diff
@@ -0,0 +1,744 @@
...
+        I::inner_product(&lhs.coeffs, rhs, zero)
+    }
+}
+
+impl<F, const DEGREE_PLUS_ONE: usize> ProjectableToField<F> for BinaryU64Poly<DEGREE_PLUS_ONE>
+where
+    F: PrimeField + FromRef<F> + 'static,
+{
+    #![allow(clippy::arithmetic_side_effects)] // False alert, field operations are safe
+    fn prepare_projection(sampled_value: &F) -> impl Fn(&Self) -> F + 'static {
```
frozenspider, line 385: Extract common code with `binary_ref`?

frozenspider (changes requested):

osdnk:
@frozenspider Your comments should be addressed

poly/src/univariate/binary_shared.rs
```diff
@@ -0,0 +1,34 @@
+use crypto_primitives::PrimeField;
+use zinc_utils::from_ref::FromRef;
+
+/// Generic projection helper for binary polynomials.
+/// `get_coeff` should return true if the i-th coefficient is 1.
+pub fn prepare_projection<F, P, GetCoeff, const N: usize>(
```
frozenspider: Nit: This can be placed in `univariate.rs` itself instead of creating a new module. In this case, this function can be made private, as child modules have access to their direct parents' private members
(Otherwise, it should be `pub(crate)`)

poly/src/univariate/binary_u64.rs
```diff
@@ -0,0 +1,752 @@
...
+use zinc_utils::{
+    from_ref::FromRef,
+    inner_product::{InnerProduct, InnerProductError},
+    mul_by_scalar::WideningMulByScalar,
+    named::Named,
+    projectable_to_field::ProjectableToField,
+};
+
+#[derive(
+    AddAssign,
```
frozenspider: Does deriving `AddAssign`, `MulAssign`, `SubAssign`, `Product` do anything, given that we defined them manually?

poly/src/univariate/binary_u64.rs
```diff
@@ -0,0 +1,752 @@
...
+            false.into(),
+            false.into(),
+            false.into(),
+            true.into(),
+            true.into(),
+            true.into(),
+            false.into(),
+            true.into(),
+            false.into(),
+        ];
```
frozenspider: That's a bit ugly 🙂 Could we do `[1, 0, 1, 1, ...].into_iter().map(|x| x != 0).collect_vec()` ?

frozenspider (approved):
Good job! Left a few minor comments, otherwise LGTM

poly/src/univariate/binary_u64.rs
```diff
@@ -0,0 +1,752 @@
...
+    #[inline(always)]
+    fn inner_product(
+        lhs: &BinaryU64Poly<DEGREE_PLUS_ONE>,
+        rhs: &[Rhs],
+        zero: Out,
+    ) -> Result<Out, InnerProductError> {
+        let lhs =
+            DensePolynomial::<Boolean, DEGREE_PLUS_ONE>::new(
+                array::from_fn::<_, DEGREE_PLUS_ONE, _>(|i| Boolean::new((lhs.0 & (1 << i)) != 0))
+                    as [Boolean; DEGREE_PLUS_ONE],
```
frozenspider: I think `as [Boolean; DEGREE_PLUS_ONE]` can be removed without causing errors

poly/src/univariate/binary_u64.rs
osdnk: ah, no sorry, forgot to delete

PR #71, opened 2026-01-26
Add `Euclid` trait for `DensePolynomial`

amit0365:
This PR adds the Euclid trait for DensePolynomial, implementing Euclidean division (div_rem_euclid, div_euclid, rem_euclid). This is needed to check polynomial membership in an ideal.

### DIFF
--- a/poly/src/univariate/dynamic/over_field.rs
+++ b/poly/src/univariate/dynamic/over_field.rs
@@ -1,9 +1,9 @@
 use crypto_primitives::{PrimeField, Semiring};
 use derive_more::From;
-use num_traits::{CheckedAdd, CheckedMul, CheckedNeg, CheckedSub, ConstZero, Zero};
+use num_traits::{CheckedAdd, CheckedMul, CheckedNeg, CheckedSub, ConstZero, Euclid, Zero};
 use std::{
     fmt::Display,
-    ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign},
+    ops::{Add, AddAssign, Div, Mul, MulAssign, Neg, Rem, Sub, SubAssign},
 };
 use zinc_utils::UNCHECKED;
 
@@ -243,6 +243,89 @@ impl<F: PrimeField> MulAssign<&Self> for DynamicPolynomialF<F> {
 
 impl<F: PrimeField> Semiring for DynamicPolynomialF<F> {}
 
+impl<F: PrimeField> DynamicPolynomialF<F> {
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn div_rem(&self, divisor: &Self) -> (Self, Self) {
+        let zero_poly = DynamicPolynomialF::zero();
+
+        assert!(!self.coeffs.is_empty(), "dividend is empty");
+
+        let cfg = self.coeffs[0].cfg();
+        let zero = F::zero_with_cfg(cfg);
+
+        let divisor_deg = divisor.degree().expect("division by zero polynomial");
+
+        let dividend_deg = self
+            .degree()
+            .expect("dividend has non-empty coeffs but all are zero");
+
+        if dividend_deg < divisor_deg {
+            return (zero_poly.clone(), self.clone());
+        }
+
+        // Copy dividend as working remainder
+        let mut remainder = self.coeffs.clone();
+        let quotient_len = dividend_deg - divisor_deg + 1;
+        let mut quotient = vec![zero.clone(); quotient_len];
+
+        let divisor_lead_inv = F::one_with_cfg(cfg) / divisor.coeffs[divisor_deg].clone();
+
+        for i in (0..quotient_len).rev() {
+            let rem_idx = i + divisor_deg;
+            if remainder[rem_idx] == zero {
+                continue;
+            }
+
+            // Compute quotient coefficient
+            let coeff = remainder[rem_idx].clone() * &divisor_lead_inv;
+
+            // Subtract coeff * x^i * divisor from remainder (skip leading term)
+            for j in 0..divisor_deg {
+                remainder[i + j] -= coeff.clone() * &divisor.coeffs[j];
+            }
+            // Leading coefficient is guaranteed to be zero
+            remainder[rem_idx] = zero.clone();
+
+            quotient[i] = coeff;
+        }
+
+        (
+            DynamicPolynomialF { coeffs: quotient },
+            DynamicPolynomialF { coeffs: remainder },
+        )
+    }
+}
+
+impl<F: PrimeField> Div for DynamicPolynomialF<F> {
+    type Output = Self;
+
+    fn div(self, rhs: Self) -> Self::Output {
+        self.div_rem(&rhs).0
+    }
+}
+
+impl<F: PrimeField> Rem for DynamicPolynomialF<F> {
+    type Output = Self;
+
+    fn rem(self, rhs: Self) -> Self::Output {
+        self.div_rem(&rhs).1
+    }
+}
+
+impl<F: PrimeField> Euclid for DynamicPolynomialF<F> {
+    fn div_euclid(&self, v: &Self) -> Self {
+        self.div_rem(v).0
+    }
+
+    fn rem_euclid(&self, v: &Self) -> Self {
+        self.div_rem(v).1
+    }
+
+    fn div_rem_euclid(&self, v: &Self) -> (Self, Self) {
+        self.div_rem(v)
+    }
+}
+
 impl<F: PrimeField, const DEGFEE_PLUS_ONE: usize> From<DensePolynomial<F, DEGFEE_PLUS_ONE>>
     for DynamicPolynomialF<F>
 {
@@ -278,9 +361,12 @@ impl<F: PrimeField> EvaluatablePolynomial<F, F> for DynamicPolynomialF<F> {
 }
 
 #[cfg(test)]
+#[allow(clippy::arithmetic_side_effects)]
 mod tests {
     use crypto_bigint::{Odd, modular::MontyParams};
     use crypto_primitives::{FromWithConfig, crypto_bigint_monty::F256};
+    use proptest::prelude::*;
+    use rand::Rng;
 
     use super::*;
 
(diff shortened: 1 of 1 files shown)


poly/src/univariate/dense.rs
```diff
@@ -599,3 +599,275 @@ where
         I::inner_product(&lhs.coeffs, rhs, zero)
     }
 }
+
+pub trait Euclid<Divisor> {
```
frozenspider: I don't think we need a makeshift trait for this, [there's one in `num_traits` already](https://docs.rs/num/latest/num/traits/trait.Euclid.html)

poly/src/univariate/dense.rs
```diff
@@ -599,3 +599,275 @@ where
...
+    fn div_rem_euclid(&self, divisor: &DensePolynomial<R, M>) -> (Self::Quotient, Self::Remainder) {
+        let cfg = self.coeffs[0].cfg();
+        let zero = R::zero_with_cfg(cfg);
+
+        // Handle zero divisor
+        let divisor_deg = find_degree_field(&divisor.coeffs).expect("division by zero polynomial");
+
+        // Handle zero dividend
+        let Some(dividend_deg) = find_degree_field(&self.coeffs) else {
+            let zero_poly = DensePolynomial {
```
frozenspider: `DensePolynomial` already implements `Zero` trait, so you can do `DensePolynomial::zero()` instead

poly/src/univariate/dense.rs
```diff
@@ -599,3 +599,275 @@ where
...
+    }
+
+    fn rem_euclid(&self, divisor: &Divisor) -> Self::Remainder {
+        self.div_rem_euclid(divisor).1
+    }
+}
+
+/// Returns the actual degree (index of highest non-zero coefficient) for a
+/// slice of field elements. Returns `None` for the zero polynomial.
+fn find_degree_field<R: PrimeField>(coeffs: &[R]) -> Option<usize> {
```
frozenspider: Instead of being dependent on `PrimeField` (and reference a field in the name), this should be a method on `DensePolynomial` itself that accept a zero element reference

frozenspider (changes requested):

poly/src/univariate/dense.rs
```diff
@@ -599,3 +599,275 @@ where
...
+                    dividend.coeffs, reconstructed.coeffs,
+                    "Division check failed: dividend != quotient * divisor + remainder (dividend deg: {}, divisor deg: {})",
+                    a_degree, b_degree
+                );
+            }
+        }
+    }
+
+    #[test]
+    fn test_ideal_membership() {
```
ElijahVlasov: There's no need for this test to mention ideal membership. It's testing euclidean reminder and shouldn't mention other entities. Let's keep things simpler

poly/src/univariate/dense.rs
```diff
@@ -599,3 +599,275 @@ where
...
+            b: &DensePolynomial<F, ARRAY_SIZE>,
+            cfg: &MontyParams<LIMBS>,
+        ) -> DensePolynomial<F, ARRAY_SIZE> {
+            let mut result = array::from_fn(|_| F::from_with_cfg(0, cfg));
+            for (i, ai) in a.coeffs.iter().enumerate() {
+                for (j, bj) in b.coeffs.iter().enumerate() {
+                    if i + j < ARRAY_SIZE {
+                        result[i + j] = result[i + j].clone() + ai.clone() * bj;
+                    }
+                }
```
ElijahVlasov: Check out this for a better schoolbook polynomial multiplication: https://github.com/NethermindEth/zinc-plus/blob/aefa30a6201fafb8faf768d54d377fb9c9248d03/poly/src/univariate/dynamic/multiplication.rs#L6 This is based on https://en.wikipedia.org/wiki/Polynomial#Multiplication. It seems to be worth adapting that as a util func for the `zinc-poly` crate.

poly/src/univariate/dense.rs
amit0365: removed

poly/src/univariate/dense.rs
amit0365: done

poly/src/univariate/dense.rs
amit0365: I had this because euclid in num_traits expects same type for quotient and remainder. To use it we would need zero padding, but now with @ElijahVlasov Dynamic poly which uses Vec instead, I changed this to num_traits

poly/src/univariate/dense.rs
amit0365: using this

poly/src/univariate/dynamic.rs
```diff
@@ -511,6 +511,93 @@ impl<R: Ring> CheckedNeg for DynamicPolynomial<R> {
 impl<R: Ring> Semiring for DynamicPolynomial<R> {}
 impl<R: Ring> Ring for DynamicPolynomial<R> {}
 
+impl<F: PrimeField> DynamicPolynomial<F> {
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn div_rem(&self, divisor: &Self) -> (Self, Self) {
+        let zero_poly = DynamicPolynomial::zero();
+
+        if self.coeffs.is_empty() {
+            return (zero_poly.clone(), zero_poly.clone());
```
ElijahVlasov: Should we panic in this case?

poly/src/univariate/dynamic.rs
```diff
@@ -511,6 +511,93 @@ impl<R: Ring> CheckedNeg for DynamicPolynomial<R> {
...
+        }
+
+        let cfg = self.coeffs[0].cfg();
+        let zero = F::zero_with_cfg(cfg);
+
+        let divisor_deg = divisor
+            .degree_with_zero(&zero)
+            .expect("division by zero polynomial");
```
ElijahVlasov: Yeah, we definitely should for consistency

poly/src/univariate/dynamic.rs
```diff
@@ -511,6 +511,93 @@ impl<R: Ring> CheckedNeg for DynamicPolynomial<R> {
...
+
+        let Some(dividend_deg) = self.degree_with_zero(&zero) else {
+            return (zero_poly.clone(), zero_poly.clone());
```
ElijahVlasov: ditto

poly/src/univariate/dynamic.rs
```diff
@@ -511,6 +511,91 @@ impl<R: Ring> CheckedNeg for DynamicPolynomial<R> {
 impl<R: Ring> Semiring for DynamicPolynomial<R> {}
 impl<R: Ring> Ring for DynamicPolynomial<R> {}
 
+impl<F: PrimeField> DynamicPolynomial<F> {
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn div_rem(&self, divisor: &Self) -> (Self, Self) {
+        let zero_poly = DynamicPolynomial::zero();
+
+        assert!(!self.coeffs.is_empty(), "dividend is empty");
```
ElijahVlasov: empty is synonymous to zero

amit0365: The dividend can be zero - zero length or have all coefficients equal to zero. Dont we need to check for both? Dynamic poly creates zero length but a dividend can became zero after some computation

poly/src/univariate/dynamic.rs
```diff
@@ -849,4 +935,106 @@ mod tests {
...
+                reconstructed_trimmed.trim_with_zero(&zero);
+
+                assert_eq!(
+                    dividend_trimmed, reconstructed_trimmed,
+                    "Division check failed: dividend != quotient * divisor + remainder (dividend deg: {}, divisor deg: {})",
+                    a_degree, b_degree
+                );
+            }
+        }
+    }
```
ElijahVlasov: This could be better done with `proptest`

amit0365: Didnt know about this. Changing to proptest

ElijahVlasov (approved):
LGTM!

amit0365:
@codex review

poly/src/univariate/dynamic/over_field.rs
```diff
@@ -243,6 +243,89 @@ impl<F: PrimeField> MulAssign<&Self> for DynamicPolynomialF<F> {
...
 impl<F: PrimeField> Semiring for DynamicPolynomialF<F> {}
 
+impl<F: PrimeField> DynamicPolynomialF<F> {
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn div_rem(&self, divisor: &Self) -> (Self, Self) {
+        let zero_poly = DynamicPolynomialF::zero();
+
+        assert!(!self.coeffs.is_empty(), "dividend is empty");
+
+        let cfg = self.coeffs[0].cfg();
```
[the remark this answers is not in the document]
amit0365: @frozenspider @ElijahVlasov we should only panic when divisor is zero no?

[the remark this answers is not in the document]
ElijahVlasov: Yes, if the dividend is empty it means it is zero, that's fine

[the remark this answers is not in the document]
frozenspider: +1 to returning (zero, zero) as this seems sound. Dividing a zero polynomial is a valid operation.
Also, we should probably use `self.degree()` instead of checking `coeffs` length - it might be untrimmed.

poly/src/univariate/dynamic/over_field.rs
```diff
@@ -243,6 +243,89 @@ impl<F: PrimeField> MulAssign<&Self> for DynamicPolynomialF<F> {
...
+    fn div_euclid(&self, v: &Self) -> Self {
+        self.div_rem(v).0
+    }
+
+    fn rem_euclid(&self, v: &Self) -> Self {
+        self.div_rem(v).1
+    }
+
+    fn div_rem_euclid(&self, v: &Self) -> (Self, Self) {
+        self.div_rem(v)
```
frozenspider: Why do we need a separate `div_rem` method if there's a dedicated `div_rem_euclid`?

poly/src/univariate/dynamic/over_field.rs
```diff
@@ -243,6 +243,89 @@ impl<F: PrimeField> MulAssign<&Self> for DynamicPolynomialF<F> {
...
+    type Output = Self;
+
+    fn rem(self, rhs: Self) -> Self::Output {
+        self.div_rem(&rhs).1
+    }
+}
+
+impl<F: PrimeField> Euclid for DynamicPolynomialF<F> {
+    fn div_euclid(&self, v: &Self) -> Self {
+        self.div_rem(v).0
```
frozenspider: Not needed, default implementation does the same

poly/src/univariate/dynamic/over_field.rs
```diff
@@ -243,6 +243,89 @@ impl<F: PrimeField> MulAssign<&Self> for DynamicPolynomialF<F> {
...
+    }
+
+    fn rem_euclid(&self, v: &Self) -> Self {
+        self.div_rem(v).1
```
frozenspider: Ditto

frozenspider (changes requested):

poly/src/univariate/dynamic/over_field.rs
amit0365: Which default impl are you referring to? I think only [div_rem_euclid](https://docs.rs/num/latest/num/traits/trait.Euclid.html#method.div_rem_euclid) has default impl which calls div_euclid and rem_euclid. Instead we can compute both quotient and remainder in div_rem_euclid and have div_euclid and rem_euclid call it.

frozenspider: Oh sorry, you're right, my bad! No idea what code I was looking at.

poly/src/univariate/dynamic/over_field.rs
amit0365: Removed

frozenspider (approved):
LGTM!

PR #72, opened 2026-01-27
Introduce dynamic polynomials

ElijahVlasov:
Zinc+ PIOP in general doesn't require polynomials to observe the degree bounds introduced by Zip+. That makes it difficult  to define UAIR in a way that would allow using only `DensePolynomial` and at the same time allow multiplication of witness cells.

This PR introduces `DynamicPolynomialF` and `DynamicPolynomialFS` types for polynomials of arbitrary degree over a field and a fixed semiring respectively. The coefficients of the polynomials are stored in vectors.


poly/src/univariate/dynamic.rs
```diff
@@ -0,0 +1,852 @@
...
+
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn trim(&mut self) {
+        self.coeffs
+            .truncate(self.degree().map_or(0, |degree| degree + 1));
+    }
+}
+
+impl<R> DynamicPolynomial<R> {
+    pub const fn zero() -> Self {
```
frozenspider: I don't think we need this given that we have `ConstZero` trait

poly/src/univariate/dynamic.rs
```diff
@@ -0,0 +1,852 @@
...
+    }
+
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn trim_with_zero(&mut self, zero: &R) {
+        self.coeffs
+            .truncate(self.degree_with_zero(zero).map_or(0, |degree| degree + 1));
+    }
+}
+
+impl<R> Default for DynamicPolynomial<R> {
```
frozenspider: Can be derived

poly/src/univariate/dynamic.rs
```diff
@@ -0,0 +1,852 @@
...
+        }
+    }
+}
+
+impl<R: Ring> Sub for DynamicPolynomial<R> {
+    type Output = Self;
+
+    #[allow(clippy::arithmetic_side_effects)]
+    fn sub(mut self, mut rhs: Self) -> Self::Output {
+        if self.coeffs.len() < rhs.coeffs.len() {
```
frozenspider: (Applies to this and the rest of operators)
I think `self.coeffs.resize(max(self.coeffs.len(), rhs.coeffs.len()), R::zero());` would be more ergonomic and won't worsen performance, if the bound `R: FixedRing` fits us

poly/src/univariate/dynamic.rs
```diff
@@ -0,0 +1,852 @@
...
+                });
+        }
+    }
+}
+
+impl<R: Semiring> AddAssign<&Self> for DynamicPolynomial<R> {
+    #[allow(clippy::arithmetic_side_effects)] // by definition
+    fn add_assign(&mut self, rhs: &Self) {
+        if self.coeffs.len() < rhs.coeffs.len() {
+            self.add_assign(rhs.clone());
```
frozenspider: (Here and in other assign ops)
Maybe we should make `AddAssign<&Self>` the main implementations so that we won't have to unconditionally clone rhs.
Also, note about coeffs resize still applies.

poly/src/univariate/dynamic.rs
```diff
@@ -0,0 +1,852 @@
...
+        // capacity.
+        unsafe {
+            coeffs.set_len(degree + 1);
+        }
+
+        Some(Self::Output { coeffs })
+    }
+}
+
+impl<R: Ring> Neg for DynamicPolynomial<R> {
```
frozenspider: Nit: Could you please keep impl order the same as in primitives? It's weird to see `Neg` so far away from `Add/Sub/Mul` and in-between `Checked*` ops

poly/src/univariate/dynamic/multiplication.rs
```diff
@@ -0,0 +1,54 @@
+use std::mem::MaybeUninit;
+
+use crypto_primitives::Semiring;
+
+#[allow(clippy::arithmetic_side_effects)]
+pub(crate) fn mul_schoolbook_unchecked<R: Semiring>(
```
frozenspider: Now that #73 is merging in, use `<const CHECK: bool>` bound and unite those two?

frozenspider (changes requested):

poly/src/univariate/dynamic.rs
ElijahVlasov: That's the problem. `R: FixedRing` doesn't satisfy us since we need dynamic polynomials over random fields specifically

frozenspider: Hmm, if we are only interested in fiedls (i.e. `F: PrimeField`), we could use `zero_with_cfg` instead

ElijahVlasov: But then we get annoying corner cases like what if we sub two empty polys and we have nowhere to draw a field config from

Uh, ignore this. We don't need a config then

ElijahVlasov: Yeah, let me rethink this whole PR a bit with an assumption we ever only deal with fields

poly/src/univariate/dynamic/multiplication.rs
```diff
@@ -0,0 +1,42 @@
+use std::mem::MaybeUninit;
+
+use crypto_primitives::Semiring;
+
+#[allow(clippy::arithmetic_side_effects)]
+pub(crate) fn mul_schoolbook<R: Semiring, const CHECKED: bool>(
```
frozenspider: I would prefer `CHECK` parameter and `CHECKED/UNCHECKED` param values

poly/src/univariate/dynamic/multiplication.rs
```diff
@@ -0,0 +1,42 @@
...
+                    acc.checked_add(&lhs[next].checked_mul(&rhs[k - next])?)
+                })?,
+            );
+
+            Some(())
+        })?;
+
+        Some(())
+    } else {
+        out.iter_mut().enumerate().for_each(|(k, out)| {
```
frozenspider: Rust compiler is pretty good at optimizing unnecessary results and unused branches, so I would bet you could unite them and use `try_for_each`/`try_fold` in both cases without loss of performance

poly/src/univariate/dynamic/over_field.rs
```diff
@@ -0,0 +1,655 @@
...
+use zinc_utils::UNCHECKED;
+
+use crate::{
+    EvaluatablePolynomial, EvaluationError, Polynomial,
+    univariate::{dense::DensePolynomial, dynamic::multiplication::mul_schoolbook},
+};
+
+/// Polynomials of dynamic degree. The implementation
+/// is tailored to work with random finite fields.
+/// To be used in UAIF and PIOP where ZIP+ degree bound
```
frozenspider: Typo: UAIR

poly/src/univariate/dynamic/over_field.rs
```diff
@@ -0,0 +1,655 @@
...
+        }
+        write!(f, "]")?;
+        Ok(())
+    }
+}
+
+impl<F: PrimeField> Default for DynamicPolynomialF<F> {
+    fn default() -> Self {
+        Self {
+            coeffs: Default::default(),
```
frozenspider, line 68: This trivial implementation can be derived

poly/src/univariate/dynamic/over_field.rs
```diff
@@ -0,0 +1,655 @@
...
+        self.add(&rhs)
+    }
+}
+
+impl<F: PrimeField> Add<&Self> for DynamicPolynomialF<F> {
+    type Output = Self;
+
+    #[allow(clippy::arithmetic_side_effects)]
+    fn add(mut self, rhs: &Self) -> Self::Output {
+        if self.coeffs.len() < rhs.coeffs.len() {
```
frozenspider: Given that the code is identical, maybe we delegate to `AddAssign`? Same with `Sub`

frozenspider (approved):
I feel like a bunch of common code could be extracted into a common private helpers working with `&mut Vec<T>` but leaving this up to your discretion.
Left some minor comments, otherwise LGTM!

poly/src/univariate/dynamic/over_field.rs
ElijahVlasov: Sadly, the type parameter's constraint `F: PrimeField` (`PrimeField` doesn't imply `Default`) prevents it from being derived :sob:

frozenspider: Interesting! I think this is an issue, logged it as https://github.com/rust-lang/rust/issues/152062

poly/src/univariate/dynamic/over_field.rs
```diff
@@ -0,0 +1,586 @@
...
+    #[inline(always)]
+    pub fn new(coeffs: impl AsRef<[F]>) -> Self {
+        Self {
+            coeffs: Vec::from(coeffs.as_ref()),
+        }
+    }
+
+    #[inline(always)]
+    pub fn degree(&self) -> Option<usize> {
+        degree(&self.coeffs, F::is_zero)
```
frozenspider: By convention, when functions from another module are used, they aren't imported directly like structs are, but are prefixed by module name (`internals::degree`). This makes it clear that those aren't local functions.

poly/src/univariate/dynamic/internal.rs
```diff
@@ -0,0 +1,107 @@
...
+    ops::{AddAssign, Neg, SubAssign},
+};
+
+use crypto_primitives::Semiring;
+use itertools::Itertools;
+
+use crate::univariate::dynamic::multiplication::mul_schoolbook;
+
+#[allow(clippy::arithmetic_side_effects)]
+pub(crate) fn new_coeffs_trimmed<R: Clone>(coeffs: &[R], is_zero: impl Fn(&R) -> bool) -> Vec<R> {
```
frozenspider: Note that if you put those in `dynamic.rs`, you wouldn't need to make them `pub(crate)` as children can access private functions of parent, it's usually used for this exact case purpose.
Leaving this up to you though.

frozenspider (approved):

PR #73, opened 2026-01-27
Speed up Zip+ test phase

frozenspider:
Core changes:
* Changed `BinaryPolyInnerProduct` for bitsliced binary polynomial to be more efficient that just delegating to `BooleanInnerProductAdd`
* Reworked transcript: `write_const` and `write_const_many` now efficiently preallocate space needed, and they no longer allocate a separate buffer. Also added a separate `write_const_many_iter` variant.

Less important changes:
* `inner_product` and `mul_by_scalar` now accepts `const CHECK: bool` to specify whether they should check for values overflowing
  * `mul_by_twiddle` was left as-is
* As such, unified `MBSInnerProduct(Checked/Unchecked)`, `BooleanInnerProduct(Checked/Unchecked)Add` variants
* Added `CHECKED`, `UNCHECKED` constants to use as "verbose booleans" (sadly we can't use enums in constants in stable Rust yet)
* `test`/`evaluate`/`verify` phases now accept a new `CHECK_FOR_OVERFLOW` constant argument
* `combine_rows` was parallelized and reworked into a macro that takes `CHECK_FOR_OVERFLOW` into account.

Notes:
* IPRS code still use checked `mul_by_scalar` to not accidentally alter its behaviour

Benchmark changes for 32-bit binary polynomial:
| Variables | Before, single-threaded | After, single-threaded | Before, parallel | After, parallel |
| --------- | ----------------------- | ---------------------- | ---------------- | --------------- |
| 2^12      | 1.3 ms                  | 0.6 ms                 | 3.1 ms           | 0.6 ms          |
| 2^16      | 20.6 ms                 | 6 ms                   | 52.3 ms          | 4.6 ms          |


poly/src/univariate/binary_u64.rs
```diff
@@ -356,25 +356,42 @@ impl<const DEGREE_PLUS_ONE: usize> From<&BinaryU64Poly<DEGREE_PLUS_ONE>>
...
+                rhs: rhs.len(),
+            });
+        }
+
+        let mut acc = zero;
+        let mut bits = lhs.0;
+        while bits != 0 {
+            let i = bits.trailing_zeros() as usize;
+            let rhs = Out::from_ref(&rhs[i]);
+            if CHECK {
```
osdnk, line 386: Question: Since this is const-generic, I presume it doesn't cause branching, and the predictor works fine?

poly/src/univariate/binary_u64.rs
```diff
@@ -356,25 +356,42 @@ impl<const DEGREE_PLUS_ONE: usize> From<&BinaryU64Poly<DEGREE_PLUS_ONE>>
...
+                acc = acc.checked_add(&rhs).ok_or(InnerProductError::Overflow)?;
+            } else {
+                acc = acc + rhs;
+            }
+            bits &= bits - 1;
```
osdnk, line 392:
```suggestion
            // changes the LSB 1 bit to 0  
            bits &= bits - 1;
```

I spent a moment thinking

poly/src/univariate/binary_u64.rs
```diff
@@ -356,25 +356,42 @@ impl<const DEGREE_PLUS_ONE: usize> From<&BinaryU64Poly<DEGREE_PLUS_ONE>>
...
+                lhs: DEGREE_PLUS_ONE,
+                rhs: rhs.len(),
+            });
+        }
+
+        let mut acc = zero;
+        let mut bits = lhs.0;
+        while bits != 0 {
+            let i = bits.trailing_zeros() as usize;
+            let rhs = Out::from_ref(&rhs[i]);
```
osdnk, line 385:
```suggestion
            let rhs = Out::from_ref(unsafe { rhs.get_unchecked(i) });
```

You checked it already above, so it seems unnecessary

zip-plus/benches/zip_benches.rs
```diff
@@ -30,21 +31,21 @@ impl ZipTypes for BenchZipTypes {
...
-    type ArrCombRDotChal = MBSInnerProductUnchecked;
+    type ArrCombRDotChal = MBSInnerProduct;
 }
 
 #[derive(Clone, Copy)]
 struct BenchRaaConfig;
 impl RaaConfig for BenchRaaConfig {
     const PERMUTE_IN_PLACE: bool = false;
-    const CHECK_FOR_OVERFLOWS: bool = false;
+    const CHECK_FOR_OVERFLOWS: bool = UNCHECKED;
```
osdnk, line 41: Is there any reasoning why in config it's `CHECK_FOR_OVERFLOWS` and otherwise just `CHECK`

zip-plus/benches/zip_plus_benches.rs
```diff
@@ -52,27 +48,21 @@ where
     type Cw = DensePolynomial<CwCoeff, D_PLUS_ONE>;
     type Fmod = Uint<{ INT_LIMBS * 4 }>;
     type PrimeTest = MillerRabin;
-    type Chal = i128;
+    type Chal = Int<{ INT_LIMBS * 2 }>;
```
osdnk: How significant is that change?

utils/src/lib.rs
```diff
@@ -8,6 +8,10 @@ pub mod ops_macros;
 pub mod parallel;
 pub mod projectable_to_field;
 
+// Can't use enums in consts in stable Rust yet, so we use consts instead.
```
osdnk:
```suggestion
// Can't use enums in const generics in stable Rust yet, so we use consts instead.
```

Once I saw this pattern, but idk, just for reference

```rust
use core::marker::PhantomData;

pub trait Mode {
    const ID: u8;
}

pub struct Checked;
pub struct Unchecked;

impl Mode for Checked { const ID: u8 = 0; }
impl Mode for Unchecked { const ID: u8 = 1; }

pub struct Foo<M: Mode>(PhantomData<M>);
```

osdnk (commented):
Some small comments and questions. Code looks great.

zip-plus/src/code/iprs.rs
```diff
@@ -54,7 +58,7 @@ where
             Config::INPUT_LEN
         );
 
-        pntt::radix8::pntt::<_, _, _, M, MBSMulByTwiddle>(row, &self.pntt_params)
+        pntt::radix8::pntt::<_, _, _, M, MBSMulByTwiddle<CHECKED>>(row, &self.pntt_params)
```
ElijahVlasov, line 61: It's safe to have a switch between `CHECKED` and `UNCHECKED` for IPRS too. Moreover, I believe switching to unchecked in IPRS will be a pretty good performance gain

poly/src/univariate/binary_u64.rs
frozenspider: Yep: https://godbolt.org/z/xsx4djhf7

poly/src/univariate/binary_u64.rs
frozenspider: Tried that, and it does not give any visible speedup. As such, I'd prefer to keep things safe.

utils/src/lib.rs
frozenspider: Interesting! Feels a little weird though 🤔

zip-plus/benches/zip_benches.rs
frozenspider: I was thinking we were still checking e.g. bit sizes to ensure they fit, but I guess those are sanity checks and are pretty much free so should be performed anyway.
How would you call this value globally? `CHECK` sounds a little vague when used outside of a specific operation

zip-plus/benches/zip_plus_benches.rs
frozenspider: Not significant at all tbh, at least the effect is not visible compared to natural fluctuations of our benches. I should probably roll this back

zip-plus/src/code/iprs.rs
frozenspider: Can we really be sure there's no chance of overflow happening, i.e. can we prove that?

ElijahVlasov: For certain params, yes, we can prove. I'm not saying let's have it always unchecked I suggest we have a flag as everywhere that controls that behaviour

frozenspider: Looked into it, I'd prefer to do it separately, so created #75 as a follow-up. Maybe also makes sense to think of making `add!`, etc. macros configurable as well

ElijahVlasov: Sounds good to me

zip-plus/src/pcs_transcript.rs
```diff
@@ -110,32 +109,47 @@ impl PcsTranscript {
...
         I: IntoIterator<Item = &'a T>,
     {
-        let mut buf = vec![0; T::NUM_BYTES];
-        for v in vs {
-            self.write_const(v, &mut buf)?;
-        }
+        let inner = self.stream.get_mut();
+        let old_len = inner.len();
+        let new_len = add!(old_len, mul!(vs_len, T::NUM_BYTES));
+        inner.resize(new_len, 0_u8);
```
ElijahVlasov, line 140: Would it make sense to provide the capacity for the  `PcsTranscript` constructor. We always know (or can estimate) how much we need so that resizes don't lead to reallocs. Feel free to leave a todo.
https://github.com/NethermindEth/zinc-plus/blob/c00cf5c3234d0031b46965848520d046eda58c6d/zip-plus/src/pcs_transcript.rs#L27

ElijahVlasov (approved):
Amazing work! LGTM!

zip-plus/src/pcs_transcript.rs
frozenspider: Good call! I think we could potentially win another 5-10% on this.
Turned out to be slightly less trivial than I thought, so made #76 to follow this up

zip-plus/benches/zip_benches.rs
frozenspider: @osdnk I'm going to merge this as-is but feel free to leave a comment and we can always follow this up

PR #78, opened 2026-02-02
UAIR and Ideal Check

ElijahVlasov:
This PR introduces the following concepts:

1. Universal AIRs. The approach is similar to modern zk circuit description frameworks, e.g. Plonky3. An implementation of a `Uair` trait receives current and next rows of "symbolic" expressions and is allowed to perform `Semiring` operations on them. This allows to abstract away constraint building logic and reuse `Uair` implementations in different contexts (combined polynomial evaluation, ideal collection, constraint counting).
2. Ideals and Ideal Checks. In `Uair`s we will encounter different types of ideals and we should be able to implement different logic of ideal checking depending on the ideals, e.g. zero ideals, ideals of the form `X^w-a`, `X-a`, general ideals generated by arbitrary polynomials.
3. Ideal check subprotocol. The subprotocol of zinc+ in which the prover sends mle evaluations of combined polynomials and the verifier checks that those belong to the ideals.


transcript/src/traits.rs
```diff
@@ -86,6 +86,15 @@ pub trait Transcript {
         F::new_with_cfg(random_inner, cfg)
     }
 
+    /// Generates a pseudorandom transcribable values as challenges based on the
+    /// current transcript state, updating it.
+    fn get_field_challenges<F: PrimeField>(&mut self, n: usize, cfg: &F::Config) -> Vec<F>
+    where
+        F::Inner: ConstTranscribable,
+    {
+        (0..n).map(|_| self.get_field_challenge(cfg)).collect()
```
frozenspider, line 99: Note: `get_field_challenge` is not efficient to call in a batch because each call allocates its own buffer. It might make sense to make a separate `get_challenge_with_buf` alternative to `get_challenge` if this is called with a large `n`. If not, I think this shouldn't matter much, but IMO worth leaving a TODO note.

uair/src/ideal.rs
```diff
@@ -0,0 +1,74 @@
...
+    #[inline(always)]
+    fn from_ref(_ideal: &ZeroIdeal) -> Self {
+        ZeroIdeal
+    }
+}
+
+/// A dummy ideal. Convenient when ideal checks
+/// have to be ignored.
+#[derive(Clone, Copy, Debug)]
+pub struct DummyIdeal;
```
frozenspider: Nit: maybe `ImpossibleIdeal` or `EmptyIdeal`?

uair/src/ideal_collector.rs
```diff
@@ -0,0 +1,106 @@
...
+    }
+}
+
+/// A type implementing ideal trait
+/// that is used to store inner
+/// ideal type `I` but ignores all
+/// ideal checks for the sake of just
+/// collecting the inner ideals.
+#[derive(Clone, Copy, Debug)]
+pub struct CollectedIdeal<I: Ideal> {
```
frozenspider: I'm not sure `CollectedIdeal` is a good name for this.
Maybe something like
```rust
pub enum IdealOrZero<I: Ideal> {
    Ideal(I),
    Zero
}
```
?

Also, the comment is now outdated.

uair/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    /// Number of witness columns the `Uair` is supposed to have.
+    fn num_cols() -> usize;
+
+    /// A general method for describing constraints.
+    ///
+    /// # Arguments
+    /// - `b`: a builder encapsulating the constraint storing logic. Its type
+    ///   `B` has to have compatible `B::Ideal` with the `Self::Ideal`, i.e. it
+    ///   must implement `FromRef<Self::Ideal>` trait.
```
frozenspider, line 55: Do we expect `Self::Ideal` to be different from `B::Ideal`?

poly/src/univariate/ideal.rs
```diff
@@ -0,0 +1,62 @@
...
+use zinc_uair::{
+    ideal::{Ideal, IdealCheck},
+    ideal_collector::CollectedIdeal,
+};
+use zinc_utils::from_ref::FromRef;
+
+use crate::{EvaluatablePolynomial, univariate::dynamic::over_field::DynamicPolynomialF};
+
+#[derive(Clone, Copy, Debug)]
+pub struct DegreeOneIdeal<R: Semiring> {
```
frozenspider, line 16: Could you please add a rustdoc comment of what that thing is mathematically?

piop/src/ideal_check.rs
```diff
@@ -0,0 +1,315 @@
...
+    /// # Parameters
+    /// - `transcript`: the Fiat-Shamir transcript.
+    /// - `trace`: the input trace for the UAIR `U`.
+    /// - `num_constraints`: the number of constraints the UAIR `U` encodes.
+    /// - `num_vars`: the number of variables the trace row MLEs have.
+    /// - `field_cfg`: random field configuration sampled on the previous steps
+    ///   of the overall protocol.
+    #[allow(clippy::type_complexity)]
+    pub fn prove_as_subprotocol<U>(
+        transcript: &mut impl Transcript,
```
frozenspider, line 57: This seems to always work with a fresh transcript. If so, maybe create one inside and not take it as an input?
Or is it done this way for consistency with `RFSumcheck` and `MLSumcheck`? (Note that latter also doesn't seem to need a transcript input)

piop/src/ideal_check.rs
```diff
@@ -0,0 +1,315 @@
...
+            num_constraints,
+        );
+        let mut transcription_buf: Vec<u8> = vec![0; <IcTypes::F as Field>::Inner::NUM_BYTES];
+
+        let evaluation_point = transcript.get_field_challenges(num_vars, field_cfg);
+
+        let combined_mle_values = cfg_iter!(combined_mles)
+            .map(|combined_mle| {
+                Ok(DynamicPolynomialF::new_trimmed(
+                    cfg_iter!(combined_mle)
```
frozenspider, line 88: I don't think it makes sense to parallelize inner workflow in addition to outer, could you check if that improves performance? (In certain cases this might make running time worse)

piop/src/ideal_check.rs
```diff
@@ -0,0 +1,315 @@
...
+
+        let evaluation_point = transcript.get_field_challenges(num_vars, field_cfg);
+
+        for mle_value in &combined_mle_values {
+            transcript.absorb_random_field_slice(&mle_value.coeffs, &mut transcription_buf);
+        }
+
+        let ideal_collector = collect_ideals::<_, U>(num_constraints);
+
+        batched_ideal_check::<_, _>(
```
frozenspider: Nit: If we call it with all-placeholder types, this means we don't need `<_, _>` at all

piop/src/ideal_check/combined_poly_builder.rs
```diff
@@ -0,0 +1,196 @@
...
+use zinc_uair::{ConstraintBuilder, Uair, ideal::DummyIdeal};
+use zinc_utils::from_ref::FromRef;
+
+use crate::ideal_check::structs::IdealCheckTypes;
+
+/// Given a UAIR `U` and a trace `trace` this function
+/// obtains the combined polynomials' MLE coefficients.
+/// Since each coefficient is also a univariate polynomial
+/// we split the resulting MLE into coefficient MLEs.
+#[allow(clippy::arithmetic_side_effects)]
```
frozenspider: Nit: This is not needed, there's no arithmetic ops with potential overflows happening

piop/src/ideal_check/combined_poly_builder.rs
```diff
@@ -0,0 +1,196 @@
...
+    let num_cols = trace.len();
+
+    let trace_matrix =
+        project_trace_matrix::<IcTypes, _>(num_rows, num_cols, trace, projecting_element);
+
+    let field_zero = IcTypes::F::zero_with_cfg(projecting_element.cfg());
+
+    let mut max_degrees_and_combined_poly_rows: Vec<(usize, Vec<DynamicPolynomialF<IcTypes::F>>)> =
+        trace_matrix
+            .as_rows()
```
frozenspider, line 41: Does it make sense to parallelize this?

piop/src/ideal_check/combined_poly_builder.rs
```diff
@@ -0,0 +1,196 @@
...
+
+pub struct CombinedPolyRowBuilder<F: PrimeField> {
+    combined_evaluations: Vec<DynamicPolynomialF<F>>,
+}
+
+impl<F: PrimeField> ConstraintBuilder for CombinedPolyRowBuilder<F> {
+    type Expr = DynamicPolynomialF<F>;
+    type Ideal = DummyIdeal;
+
+    #[allow(clippy::arithmetic_side_effects)]
```
frozenspider: Nit: This is not needed, there's no arithmetic ops with potential overflows happening

piop/src/ideal_check/combined_poly_builder.rs
```diff
@@ -0,0 +1,196 @@
...
+                combine_rows_and_get_max_degree::<IcTypes, U, _>(
+                    up,
+                    down,
+                    num_constraints,
+                    projecting_element,
+                )
+            })
+            .collect();
+
+    let max_degree = *max(max_degrees_and_combined_poly_rows
```
frozenspider: Nit: A bit more readable would be to use `x.map(...).max()` as opposed to `max(x.map(...))`

piop/src/ideal_check/combined_poly_builder.rs
```diff
@@ -0,0 +1,196 @@
...
+        |x| x.project_coefficients(projecting_element).into(),
+        |x, y| Some(DynamicPolynomialF::from(y.project_coefficients(projecting_element)) * x),
+        DummyIdeal::from_ref,
+    );
+
+    let mut combined_evaluations = constraint_builder.combined_evaluations;
+
+    combined_evaluations.iter_mut().for_each(|eval| eval.trim());
+
+    let max_degree = max(combined_evaluations
```
frozenspider: Ditto about `max()`

piop/src/ideal_check/combined_poly_builder.rs
```diff
@@ -0,0 +1,196 @@
...
+                    max_degrees_and_combined_poly_rows
+                        .iter()
+                        .map(|(_, row)| {
+                            if coeff >= row[constraint].coeffs.len() {
+                                field_zero.inner().clone()
+                            } else {
+                                row[constraint].coeffs[coeff].inner().clone()
+                            }
+                        })
+                        .collect_dense_mle_with_zero(field_zero.inner())
```
frozenspider, line 166: Nit: Given that `collect_dense_mle_with_zero` is only used here, we might want to consider removing the trait and just doing it inline

piop/src/ideal_check/combined_poly_builder.rs
```diff
@@ -0,0 +1,196 @@
...
+fn prepare_coefficient_mles<
+    IcTypes: IdealCheckTypes<DEGREE_PLUS_ONE>,
+    const DEGREE_PLUS_ONE: usize,
+>(
+    num_constraints: usize,
+    max_degree: usize,
+    max_degrees_and_combined_poly_rows: &[(usize, Vec<DynamicPolynomialF<IcTypes::F>>)],
+    field_zero: &IcTypes::F,
+) -> Vec<Vec<DenseMultilinearExtension<<IcTypes::F as Field>::Inner>>> {
+    (0..num_constraints)
```
frozenspider, line 153: Again, do you think it might make sense to parallelize this, or is it way to fast to bother?

frozenspider (changes requested):
Great job @ElijahVlasov! Left some comments, mostly minor and style

piop/src/ideal_check.rs
ElijahVlasov: All `prove_as_subprotocol` funcs are not expected to have a particular place in the sequence of subprotocols. Ideal check is not going to be the first subprotocol anyway. Commitments to the witness are done prior to it, so it should receive the FS transcript after commitments.

ElijahVlasov: EDIT: rather _they_ do not expect to be placed somewhere

piop/src/ideal_check/combined_poly_builder.rs
ElijahVlasov: I'm gonna use it a lot in the next stages of the protocol :)

frozenspider (approved):

PR #79, opened 2026-02-04
Speed up Zip+ verify phase

frozenspider:
* Parallelized verification phase.
* Optimized reading from PCS transcript by not allocating a buffer (reading directly from the transcript vector instead).
* Additionally, removed `ColumnOpening` entirely as it was obscuring the logic rather than simplifying it.

(Other endeavours to speed to verification was done by optimized limbed integer operations in NethermindEth/crypto-primitives#28 and NethermindEth/crypto-primitives#29)

Benchmark changes for 32-bit binary polynomial and IPRS codes (including changes in `crypto-primitives`):
| Variables   | Before, single-threaded | After, single-threaded | Before, parallel | After, parallel |
| ----------- | ----------------------- | ---------------------- | ---------------- | --------------- |
| 2^12 (IPRS) | 16 ms                   | 14 ms                  | 16 ms            | 7.5 ms          |
| 2^16 (IPRS) | 178 ms                  | 144 ms                 | 172 ms           | 35  ms          |
| 2^12 (RAA)  | 40 ms                   | 31 ms                  | 40 ms            | 11 ms           |
| 2^16 (RAA)  | 145 ms                  | 113 ms                 | 151 ms           | 28  ms          |

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,7 @@ license = "Apache-2.0"
 publish = false
 
 [workspace.dependencies]
-#crypto-primitives = { git = "https://github.com/NethermindEth/crypto-primitives.git", branch = "main", features = ["std", "crypto_bigint", "rand"] }
-crypto-primitives = { path = "../crypto-primitives", features = ["std", "crypto_bigint", "rand"] }
+crypto-primitives = { git = "https://github.com/NethermindEth/crypto-primitives.git", branch = "feature/better-multiplication", features = ["std", "crypto_bigint", "rand"] }
 
 crypto-bigint = { version = "= 0.7.0-rc.9", features = ["zeroize"] }
 criterion = "0.7.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ license = "Apache-2.0"
 publish = false
 
 [workspace.dependencies]
-crypto-primitives = { git = "https://github.com/NethermindEth/crypto-primitives.git", branch = "feature/better-multiplication", features = ["std", "crypto_bigint", "rand"] }
+crypto-primitives = { git = "https://github.com/NethermindEth/crypto-primitives.git", branch = "main", features = ["std", "crypto_bigint", "rand"] }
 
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -11,7 +11,7 @@ license = "Apache-2.0"
 publish = false
 
 [workspace.dependencies]
-crypto-primitives = { git = "https://github.com/NethermindEth/crypto-primitives.git", branch = "main", features = ["std", "crypto_bigint", "rand"] }
+crypto-primitives = { git = "https://github.com/NethermindEth/crypto-primitives.git", branch = "feature/better-multiplication", features = ["std", "crypto_bigint", "rand"] }
```
osdnk: what is the impact of this?

frozenspider: 2^16 poly, single-threaded: 460 ms -> 410 ms (before other optimizations, now it's about ~140 ms), par: 40 -> 30 ms

zip-plus/src/pcs/phase_test.rs
```diff
@@ -6,18 +6,17 @@ use crate::{
     pcs::{
         ZipPlusTestTranscript,
         structs::{ZipPlus, ZipPlusHint, ZipPlusParams, ZipTypes},
-        utils::{ColumnOpening, validate_input},
+        utils::validate_input,
     },
     pcs_transcript::PcsTranscript,
 };
 use num_traits::{ConstOne, ConstZero, Zero};
+#[cfg(feature = "parallel")]
```
osdnk, line 14: Would it make sense to add a new feature called "paraller-verifier"? I think we are somewhat interested in the case of a parallel prover / sequential verifier.

frozenspider: I think that's going to be somewhat cumbersome to implement since they both share common code, maybe for now we'd be better just making two bench runs with different feature set selected?

Cargo.toml
frozenspider: UPD: I think shift from ~400+ to ~140 was caused by lowering `NUM_COLUMN_OPENINGS` previously

ElijahVlasov (approved):
LGTM!

PR #82, opened 2026-02-06
Add unchecked addition optimization to IPRS encoding

osdnk:
Adds `const CHECK: bool` generic parameter to `IprsCode` and the underlying `PNTT` functions to allow toggling overflow checking at compile time. 
When CHECK = false (UNCHECKED), uses direct addition instead of checked_add, avoiding clone + overflow check overhead. 

```
Zip+ IPRS/EncodeRows: BPoly<31> -> Poly<i64, 31>, poly_size = 2^13
                        time:   [6.0335 ms 6.2425 ms 6.5143 ms]
                        change: [−32.360% −31.928% −29.210%] (p = 0.00 < 0.05)
                        Performance has improved.
````

### DIFF
--- a/zip-plus/src/code/iprs/pntt/radix8.rs
+++ b/zip-plus/src/code/iprs/pntt/radix8.rs
@@ -21,7 +21,7 @@ use params::*;
 pub(crate) use mul_by_twiddle::*;
 
 /// The main entrypoint of the radix-8 pseudo NTT algorithm.
-pub(crate) fn pntt<In, Out, C, MulInByTwiddle, MulOutByTwiddle, const CHECK: bool>(
+pub(crate) fn pntt<In, Out, C, MulInByTwiddle, MulOutByTwiddle, const CHECK_ADDITION: bool>(
     input: &[In],
     params: &Radix8PnttParams<C>,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: zip-plus/src/code/iprs/pntt/radix8.rs [hunk 1]
```diff
@@ -21,14 +21,21 @@ use params::*;
 pub(crate) use mul_by_twiddle::*;
 
 /// The main entrypoint of the radix-8 pseudo NTT algorithm.
-pub(crate) fn pntt<In, Out, C, MulInByTwiddle, MulOutByTwiddle>(
+pub(crate) fn pntt<In, Out, C, MulInByTwiddle, MulOutByTwiddle, const CHECK: bool>(
```
ElijahVlasov: Can we name this parameter `CHECK_ADDITION` throughout? With proliferation of `CHECK` and `CHECKED` params it's difficult to keep track which is which 😁

ElijahVlasov:
Amazing work! A small comment and g2g :slightly_smiling_face:

ElijahVlasov (approved):

PR #94, opened 2026-02-10
Combined polynomial evaluation claims resolver

### LINKED ISSUE #90: [THIS WEEK] Projection & Sumcheck - 10% to finalize the implementation (integrate with UAIR)

ElijahVlasov:
This PR introduces the second stage of the Zinc+ protocol. This is the subprotocol that reduces evaluation claims about the combined polynomial to the evaluation claims about the trace and shifted trace. 

closes https://github.com/NethermindEth/zinc-plus/issues/90


piop/src/combined_poly_resolver.rs
```diff
@@ -0,0 +1,471 @@
...
+                            .clone()
+                    })
+                    .collect()
+            })
+            .collect();
+
+        // Shifted trace. Just take the trace, drop the first row
+        // and append 0 to the end.
+        let down: Vec<DenseMultilinearExtension<F::Inner>> = cfg_iter!(up)
+            .map(|column| cfg_iter!(column[1..]).cloned().collect())
```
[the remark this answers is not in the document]
ElijahVlasov: Not that can it happen in production but let's debug assert `num_vars != 1`.

piop/src/combined_poly_resolver/structs.rs
```diff
@@ -0,0 +1,73 @@
...
+    pub down_evals: Vec<F>,
+}
+
+impl<F: PrimeField> Proof<F> {
+    /// Check if `up_evals` and `down_evals` vectors
+    /// has the length `U::num_cols()`.
+    pub fn validate_evaluation_sizes<R: Semiring + 'static, U: Uair<R>>(
+        &self,
+    ) -> Result<(), CombinedPolyResolverError<F>> {
+        let num_cols = U::num_cols();
```
frozenspider: I think it's more straightforward to ask for `num_cols` argument directly instead of adding an `<R, U>` type bound - proof doesn't seem to care about semirings and UAIR otherwise

piop/src/combined_poly_resolver.rs
```diff
@@ -0,0 +1,483 @@
...
+use std::{collections::HashMap, marker::PhantomData, slice};
+use thiserror::Error;
+use zinc_poly::{
+    EvaluatablePolynomial, EvaluationError,
+    mle::MultilinearExtensionWithConfig,
+    utils::{ArithErrors, build_eq_x_r_inner, eq_eval},
+};
+use zinc_uair::ideal::ImpossibleIdeal;
+use zinc_utils::{cfg_iter, from_ref::FromRef, powers};
+
```
frozenspider: Nit: Because of these line breaks, `cargo fmt` could not combine same-prefixed imports like `zinc_poly`, `zinc_uair`, etc.

piop/src/combined_poly_resolver.rs
```diff
@@ -0,0 +1,483 @@
...
+            .map(|column| cfg_iter!(column[1..]).cloned().collect())
+            .collect();
+
+        let eq_r = build_eq_x_r_inner(evaluation_point, field_cfg)?;
+
+        // To get the constraints on the last row ignored
+        // we multiply each constraint polynomial
+        // by the selector (1 - eq(1,...,1, x))
+        let last_row_selector =
+            build_eq_x_r_inner(&vec![F::one_with_cfg(field_cfg); num_vars], field_cfg)?;
```
frozenspider: Nit: `one.clone()`

piop/src/combined_poly_resolver.rs
```diff
@@ -0,0 +1,483 @@
...
+
+        // The challenge '\alpha' to batch multiple evaluation claims
+        let folding_challenge: F = transcript.get_field_challenge(field_cfg);
+
+        let folding_challenge_powers: Vec<F> = powers(
+            folding_challenge,
+            F::one_with_cfg(field_cfg),
```
frozenspider: Ditto

piop/src/combined_poly_resolver.rs
```diff
@@ -0,0 +1,483 @@
...
+                            .clone()
+                    })
+                    .collect()
+            })
+            .collect();
+
+        // Shifted trace. Just take the trace, drop the first row
+        // and append 0 to the end.
+        let down: Vec<DenseMultilinearExtension<F::Inner>> = cfg_iter!(up)
+            .map(|column| cfg_iter!(column[1..]).cloned().collect())
```
frozenspider: Parallelization here does nothing and is pure overhead

piop/src/combined_poly_resolver.rs
```diff
@@ -0,0 +1,483 @@
...
+                            .expect("evaluation cannot fail here")
+                            .inner()
+                            .clone()
+                    })
+                    .collect()
+            })
+            .collect();
+
+        // Shifted trace. Just take the trace, drop the first row
+        // and append 0 to the end.
```
frozenspider: Dropping the first element is obvious, but do we really append zero?

piop/src/combined_poly_resolver.rs
```diff
@@ -0,0 +1,483 @@
...
+                );
+
+                folder.folded_constraints * (one.clone() - selector) * eq_r
+            },
+            field_cfg,
+        );
+
+        // Sumcheck prover stops evaluating MLEs
+        // at the second to last challenge
+        // leaving all MLEs in num_vars=1
```
frozenspider, line 180: Do we want `debug_assert` that all MLEs do in fact have `num_vars == 1`?

piop/src/combined_poly_resolver.rs
```diff
@@ -0,0 +1,483 @@
...
+        let zero = F::zero_with_cfg(field_cfg);
+        let one = F::one_with_cfg(field_cfg);
+
+        let folding_challenge: F = transcript.get_field_challenge(field_cfg);
+
+        let folding_challenge_powers: Vec<F> =
+            powers(folding_challenge, one.clone(), num_constraints);
+
+        // Compute v_0 + \alpha * v_1 + ... + \alpha ^ k * v_k.
+        let expected_sum = ic_check_subclaim
```
frozenspider, line 262: Might make sense to parallelize this.
(This isn't as trivial as usual though because `rayon` picked different API for `try_fold` for whatever reason)

utils/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+use crypto_primitives::Semiring;
```
frozenspider: Nit: It's better to place `use` section after `mod` since it affects the following code while `mod` doesn't

frozenspider (changes requested):

piop/src/combined_poly_resolver.rs
ElijahVlasov: Leaving a todo

piop/src/combined_poly_resolver.rs
ElijahVlasov: I'm not sure if it breaks any abstraction but `from_iter` implementation for `DenseMultilinearExtension` pads the collected evaluation vector with zeros to the next pow of 2.
https://github.com/NethermindEth/zinc-plus/blob/dbf2321a8663da370594d3dd8580c8530bb5861f/poly/src/mle/dense.rs#L129

ElijahVlasov: As the AI overlord pointed out it might lead to a problem for `num_vars=1` but that's not gonna happen in practice

frozenspider: Sounds good, although I suggest we adjust a comment to reflect what's actually happening

frozenspider (approved):
Great job!

PR #96, opened 2026-02-11
`next` MLE

ElijahVlasov:
For the shifted trace sumcheck protocol we need `next` MLE construction and a means to evaluate it in log-time on the side of the verifier. 

`next` is a multilinear polynomial that selects the successor function graph. I.e., on the hypercube it takes values in `{0, 1}` and satisfies the condition:
```
next([b1], [b2]) = 1 iff b2 = b1 + 1
```

This PR adapts [this evaluation procedure implementation](https://github.com/TomWambsgans/Whirlaway/blob/9e3592b5b5c3702fe3a1728be758506a7ee9283b/crates/air/src/utils.rs#L92) and introduces a function that returns a `next` MLE as a `DenseMultilinearExtension`.


frozenspider (approved):

frozenspider:
Build fails because #135 is not yet merged, so merging this anyway

PR #102, opened 2026-02-12
New UAIR interface + decoupled projections

### LINKED ISSUE #95: [THIS WEEK] Extract projection as a separate stage

ElijahVlasov:
This PR achieves 2 things:
1. Projection logic is not a part of either ideal check or combined poly resolver subprotocols. That will allow to project binary polynomials and integers more wisely at the second stage.
2. UAIRs now have different types of witness columns. For now, we assume they are either binary polynomials, arbitrary polynomials, or integers. Scalars of UAIRs are always treated as arbitrary polynomials but in the future that should be revised too. 

The main assumption is that the e2e Zinc+ protocol takes as input something like: 
```rust
binary_poly_trace: &[DenseMultilinearExtension<BinaryPoly<DEGREE_PLUS_ONE>>],
arbitrary_poly_trace: &[DenseMultilinearExtension<DensePolynomial<PolyCoeff, DEGREE_PLUS_ONE>>],
int_trace: &[DenseMultilinearExtension<Int>],
```
which, I think, is pretty much what we want.

closes https://github.com/NethermindEth/zinc-plus/issues/95


uair/src/lib.rs
```diff
@@ -32,12 +32,76 @@ pub trait ConstraintBuilder {
...
+    pub const fn total_cols(&self) -> usize {
+        self.binary_poly_cols + self.arbitrary_poly_cols + self.int_cols
+    }
+}
+
+/// A view on a row of the trace.
+/// Contains references to cells of the trace
+/// of all types lying in the same trace row.
+#[derive(Clone, Copy)]
+pub struct TraceRow<'a, Expr> {
```
frozenspider, line 74: I see that currently we only ever use `arbitrary_poly`, everything else is not used.
However, if we intend to use all 3 of them, does it make sense for them to share a type?...

I guess this can get addressed once we actually start using them though, so we can see the usage patterns in practice.

uair/src/lib.rs
```diff
@@ -32,12 +32,76 @@ pub trait ConstraintBuilder {
...
+    /// Number of columns with arbitrary polynomial elements.
+    pub arbitrary_poly_cols: usize,
+    /// Number of columns with integers.
+    pub int_cols: usize,
+}
+
+impl UairSignature {
+    /// Maximum number of columns across the three types.
+    pub const fn max_cols(&self) -> usize {
+        // TODO(Ilia): is there a const max?
```
frozenspider: Not in baseline Rust yet, but there's one in [const-utils](https://crates.io/crates/const-utils).
However, we don't use it in const context so it being `const fn` yields us nothing.

uair/src/lib.rs
```diff
@@ -47,42 +111,49 @@ pub trait Uair<R: Semiring + 'static> {
...
     ///   convenient to provide a closure instead of a `FromRef` implementation.
     /// - `mbs`: a closure that allows to multiply expressions by `R`. Same
     ///   rationale as for `from_ref`.
-    fn constrain_general<B, FromR, MulByScalar, IFromR>(
+    fn constrain_general<'a, B, FromR, MulByScalar, IFromR>(
         b: &mut B,
-        up: &[B::Expr],
-        down: &[B::Expr],
+        up: TraceRow<'a, B::Expr>,
+        down: TraceRow<'a, B::Expr>,
```
frozenspider: Style, here and in other places: I believe we can omit an explicit lifetime `'a` altogether:
```rust
fn constrain_general<B, ...>(
    ...,
    up: TraceRow<B::Expr>,
    ...
```

poly/src/univariate/dense.rs
```diff
@@ -647,3 +635,9 @@ where
         I::inner_product::<CHECK>(&lhs.coeffs, rhs, zero)
     }
 }
+
+impl<R, const DEGREE_PLUS_ONE: usize> DensePolynomial<R, DEGREE_PLUS_ONE> {
+    pub fn map_coeffs<F: PrimeField>(&self, f: impl Fn(&R) -> F) -> DynamicPolynomialF<F> {
```
frozenspider: Do we really want an output to be dynamic-sized?
(If so, maybe call it `map_to_dyn_poly`? Otherwise it's a bit unexpected to see the output type change.)

Also this should probably be declared in a trait, given that it's shared across out polys.

piop/src/projections.rs
```diff
@@ -0,0 +1,181 @@
...
+/// using a field config. Projecting int trace is just
+/// turning constant `DynamicPolynomialF`s into `F`s.
+/// The real work may happen only for the arbitrary polynomials
+/// which we have no better solution than just evaluate them on
+/// the projecting element.
+#[allow(clippy::arithmetic_side_effects)]
+pub fn project_trace_to_field<F: PrimeField + FromRef<F> + 'static, const DEGREE_PLUS_ONE: usize>(
+    binary_poly_trace: &[DenseMultilinearExtension<BinaryPoly<DEGREE_PLUS_ONE>>],
+    arbitrary_poly_trace: &[DenseMultilinearExtension<DynamicPolynomialF<F>>],
+    int_trace: &[DenseMultilinearExtension<DynamicPolynomialF<F>>],
```
frozenspider, line 99: Do we expect inputs to this be dynamic-sized?
(Applies to other projections as well)

piop/benches/ideal_check.rs
```diff
@@ -5,83 +5,72 @@ use criterion::{
...
-    const FIELD_LIMBS: usize;
-}
-
-impl<const FIELD_LIMBS: usize, const INT_LIMBS: usize> BenchIcTrait
-    for BenchIcTypes<FIELD_LIMBS, INT_LIMBS>
-{
-    const FIELD_LIMBS: usize = FIELD_LIMBS;
-}
+type WitnessCoeff<const INT_LIMBS: usize> = Int<INT_LIMBS>;
+type Witness<const INT_LIMBS: usize> = DensePolynomial<Int<INT_LIMBS>, DEGREE_PLUS_ONE>;
```
frozenspider: Nit: `DensePolynomial<WitnessCoeff<INT_LIMBS>`

frozenspider (approved):
Overall LGTM! Left some questions but feel free to merge if you think it's better this way

test-uair/src/lib.rs
```diff
@@ -170,19 +188,24 @@ impl<const LIMBS: usize> GenerateWitness<DensePolynomial<Int<LIMBS>, 32>>
     }
 }
 
-pub struct TestAirScalarMultiplications;
+pub struct TestAirScalarMultiplications<const LIMBS: usize>;
```
frozenspider, line 193: Btw what about witness generation for it?

uair/src/lib.rs
ElijahVlasov: Added an example of a mixed-typed UAIR. Check it out! https://github.com/NethermindEth/zinc-plus/blob/2a69edc41c587f09616f15707eb578e8080b941f/test-uair/src/lib.rs#L314

piop/src/projections.rs
ElijahVlasov: Do you mean why we have `DynamicPolynomialF` there?

So this projection should happen in between of ideal check and combined poly resolver. Before ideal check we projected our trace to F[X]. I.e. we turned all coeffs of arbitrary polys into Montgomery and so did we with the int columns. ideal checks expects everything on its input as dynamics so previous projection outputs dynamics. Then we can repurpose the output of the first projection to compute the second projection. In particular it's useful for `arbitrary_poly_trace`. With `int_trace` it's a bit lame, yes, we expect those polys are necessarily constant.

test-uair/src/lib.rs
ElijahVlasov: Well, I remember creating it specifically to test scalar collection :grin: So I didn't need witness generation :upside_down_face:

poly/src/univariate/dense.rs
ElijahVlasov: With https://github.com/NethermindEth/zinc-plus/pull/110 it will be much saner

PR #110, opened 2026-02-13
Some collection traits for univariate polynomials

ElijahVlasov:
The main change here actually is `iter()` method for univariate polynomials to easily use `.map()` and other benefits of iterators with polynomials

### DIFF
--- a/poly/src/univariate/binary_ref.rs
+++ b/poly/src/univariate/binary_ref.rs
@@ -13,7 +13,7 @@ use std::{
     hash::Hash,
     iter::{Product, Sum},
     marker::PhantomData,
-    ops::{Add, AddAssign, Mul, MulAssign, Sub, SubAssign},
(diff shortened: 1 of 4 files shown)


poly/src/univariate/binary_ref.rs
```diff
@@ -316,6 +316,34 @@ impl<const DEGREE_PLUS_ONE: usize> From<&BinaryRefPoly<DEGREE_PLUS_ONE>>
     }
 }
 
+impl<const DEGREE_PLUS_ONE: usize> BinaryRefPoly<DEGREE_PLUS_ONE> {
+    #[inline(always)]
+    pub fn iter(&'_ self) -> std::slice::Iter<'_, Boolean> {
```
frozenspider: Style nit, here and in other places: no need to write `&'_ self`, it's no different than just `&self`

frozenspider (approved):

PR #116, opened 2026-02-14
feat(zip-plus): support arbitrary-shape input matrices

albert-garreta:
## Summary

This PR decouples the `zip-plus` PCS from the assumption that input matrices are square, allowing arbitrary matrix shapes (i.e. any number of rows and columns, as long as both are powers of two and their product equals the polynomial size).

## Motivation

Previously, `LinearCode::new()` received `poly_size` and internally computed `row_len = isqrt(poly_size).next_power_of_two()`, forcing a square matrix layout. This made it impossible to use non-square matrix arrangements — for example, encoding a polynomial of size $2^{15}$ into a matrix with 64 rows and 512 columns.

### Core API change
- **`LinearCode::new(poly_size)` → `LinearCode::new(row_len)`**: The trait method now takes the row length directly. Callers decide the matrix shape.

### Linear code implementations
- **`RaaCode::new()`**: No longer derives `row_len` via `isqrt()`; receives it as a parameter. The codeword bit-width calculation is updated to use `2 * row_len.ilog2() + 2 * rep_factor.ilog2()` instead of the old `num_vars`-based formula.
- **`RaaSignFlippingCode::new()`**: Delegates to `RaaCode::new(row_len)`.
- **`IprsCode::new()`**: Asserts `row_len == Config::INPUT_LEN` instead of checking `poly_size % Config::INPUT_LEN == 0`.

### PCS setup & validation
- **`ZipPlus::setup()`**: Computes `num_rows = poly_size / row_len` with assertions that both dimensions are powers of two and the division is exact — replacing the old `next_power_of_two()` rounding.
- **`validate_input()`**: Now accepts `row_len` as a parameter and computes the bit-width bound `d` from `row_len.ilog2()` rather than hard-coding `d = num_vars / 2` (the square matrix assumption).

### Call sites
- All phase files (commit, test, evaluate, verify) pass `pp.linear_code.row_len()` to `validate_input()`.
- All tests and benchmarks construct codes with `Lc::new(row_len)` instead of `Lc::new(poly_size)`.

## Backward compatibility

Existing callers that used `ZipPlus::setup(poly_size, linear_code)` continue to work — `setup()` now derives `num_rows` from `poly_size / linear_code.row_len()`. The only breaking change is in `LinearCode::new()`, which now takes `row_len` instead of `poly_size`.

## Testing

All 68 existing tests pass. No new tests were required since the existing test suite already covers the full protocol (commit → test → evaluate → verify) and exercises various matrix dimensions.

### DIFF
--- a/zip-plus/benches/zip_common.rs
+++ b/zip-plus/benches/zip_common.rs
@@ -93,8 +93,9 @@ pub fn encode_rows<Zt: ZipTypes, Lc: LinearCode<Zt>, const P: usize>(
         ),
         |b| {
             let mut rng = ThreadRng::default();
-            let poly_size = 1 << P;
-            let linear_code = Lc::new(poly_size);
+            let poly_size: usize = 1 << P;
+            let row_len = poly_size.isqrt().next_power_of_two();
+            let linear_code = Lc::new(row_len);
             let params = ZipPlus::setup(poly_size, linear_code);
             let row_len = params.linear_code.row_len();
             let poly = DenseMultilinearExtension::<<Zt as ZipTypes>::Eval>::rand(P, &mut rng);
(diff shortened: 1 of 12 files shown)


albert-garreta:
Not sure this should be a PR. My goal is to kickstart and maybe finish allowing matrices to have different sizes

PR #118, opened 2026-02-14
Optimize Merkle leaf hashing with contiguous Blake3 input (~1.6x speedup)

albert-garreta:
## Summary

Optimizes Merkle tree leaf hashing by serializing each column's data into a contiguous byte buffer before feeding it to Blake3, replacing the previous per-element `hasher.update()` approach.

## Problem

The previous `hash_leaves` implementation called `hasher.update(&buf)` once per element (8 bytes for `i64`). This prevented Blake3 from efficiently processing full 1 KiB chunks with SIMD — each tiny `update` call incurred function-call overhead, memcpy, and boundary checks.

## Solution

For each column, allocate a contiguous buffer of `num_rows * elem_bytes` bytes, serialize all elements into it, then call `hasher.update(&buf)` once. This lets Blake3 process the data in full chunks using its SIMD pipeline.

## Benchmarks

Measured using the existing Zip+ PCS benchmarks (`MerkleRoot` from `zip_common.rs` pattern):

| poly_size | Before | After | Speedup |
|-----------|--------|-------|---------|
| 2^12 (64 rows) | 197 µs | 122 µs | **1.61x** |
| 2^14 (128 rows) | 766 µs | 477 µs | **1.61x** |
| 2^16 (256 rows) | 3.10 ms | 1.95 ms | **1.59x** |

The optimization benefits all Merkle tree construction regardless of the number of rows or batch size.

## Changes

Single file changed: `zip-plus/src/merkle.rs` — `hash_leaves` function rewritten to use contiguous buffers (19 insertions, 2 deletions).

### DIFF
--- a/zip-plus/src/merkle.rs
+++ b/zip-plus/src/merkle.rs
@@ -124,14 +124,31 @@ macro_rules! hash_many {
 
 /// Construct the leaves of the Merkle tree by hashing each column across all
 /// rows.
+///
+/// For each column, serializes all row elements into a single contiguous byte
+/// buffer and feeds it to Blake3 in one `update` call.  This lets Blake3
+/// process full 1 KiB chunks with SIMD, which is significantly faster than
+/// the per-element `update` approach when columns are tall (many rows).
 fn hash_leaves<S>(rows: &[&[S]], m_cols: usize) -> Vec<MtHash>
 where
     S: ConstTranscribable + Send + Sync,
 {
+    let num_rows = rows.len();
+    let elem_bytes = S::NUM_BYTES;
+    let col_bytes = num_rows * elem_bytes;
+
     cfg_into_iter!(0..m_cols)
         .map(|i| {
-            // Hash the i-th column across all rows.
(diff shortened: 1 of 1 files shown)


albert-garreta:
Revision by a separate agent of a different brand:

-----

I reviewed the PR end-to-end and the change looks correct.

Scope is focused (single-file change in Merkle leaf hashing path).
New logic preserves serialization order/semantics while improving hashing throughput by feeding contiguous column bytes to Blake3.
Existing Merkle proof/verification compatibility appears intact.
I also ran targeted checks locally and they passed:

cargo test -p zip-plus merkle::tests::test_merkle_proof -- --nocapture
cargo test -p zip-plus pcs::phase_commit::tests::commit_with_many_variables -- --nocapture
cargo test -p zip-plus pcs::phase_verify::tests::successful_verification_of_valid_proof -- --nocapture
Non-blocking note: this trades some extra per-column buffer memory for speed, which seems reasonable for the intended optimization.

frozenspider:
Changed from here were adapted in #130

PR #128, opened 2026-02-19
Ideal checks: Use MLE-first evaluation of linear constraints

### LINKED ISSUE #103: Do not evaluate combined polys for zero-constraints in ideal-check.
### Description

1. In the `assert_zero` method of the `ConstraintBuilder` implementation for `CombinedPolyBuilder` do nothing as we don't need to deal with combined polys for constraints of the form `lhs=0` since it's zero everywhere. 
https://github.com/NethermindEth/zinc-plus/blob/c1f68ece5a127ef6312444280babde132c7b1976/piop/src/ideal_check/combined_poly_builder.rs#L176
2. Adjust the ideal-check prover/verifier accordingly.

### Benefit
Currently, we are computing combined polynomial and evaluate it on a random point for each constraint defined by a UAIR in the ideal-check subprotocol. There might be a lot of zero-constraints, i.e. constraints created via `assert_zero` of the `ConstraintBuilder`. Building and evaluating combined polynomials for those constraints is a waste of time and memory as we can assume optimistically that their combined polynomials will be zero.

### LINKED ISSUE #113: [Ideal check part] -- Performance goals
If it helps, the repo https://github.com/albert-garreta/32x-lincomb-mle-bench claims to do:

32 times in parallel:
- linear combination of m vectors of length n with entries in a fixed given 192-bit prime field F_q
- MLE evaluation of the resulting vector at a point x\in F_q^{log n}

For m=16 and n=2^8 times are of 0.25ms

Repo written, "reviewed" and "audited" by separate agents

frozenspider:
Resolves #113, also resolves #103 

Our approach for ideal check protocol prover is to compute combined polynomials by combining `up` and `down` rows first to obtain coefficient MLEs, and then evaluating them at the evaluation point.

For linear-degree constraints (those without multiplication), @ElijahVlasov proposed an alternative approach - to evaluate `up` and `down` without combining them first.

Each approach has different cost though, and heuristics on whether or not to use MLE-first evaluation for linear constraints is somewhat non-trivial (as it requires counting number of operations per row), so I'm leaving this up to the caller.

Besides this, general ideal check proving method (via combining polynomial) has been tweaked to work with transposed trace matrix, as the access pattern for MLE-first and combined polynomial approach is different.

Additionally, `IdealCheckProtocol` was changed into a trait implemented for all UAIRs, this allows for much simpler call syntax without having to specify types most of the time.

| Speedup by bench               | Single-thread | Parallelized |
| ------------------------------ | ------------- | ------------ |
| NoMult (Combined)              | ~2x           | Minor        |
| SimpleMult (Combined)          | No            | No           |
| BinaryDecomposition (Combined) | ~20%          | No           |
| BigLinearUair (Combined)       | ~50%          | ~30%         |
| BigLinearUair (MLE-first)      | ~2x           | ~2x          |


albert-garreta:
Thanks! Could we rename "direct evaluation" to "MLE-first"? I think it is more intuitive

piop/src/ideal_check/combined_poly_builder.rs
```diff
@@ -129,28 +134,151 @@ fn prepare_coefficient_mles<F: PrimeField>(
...
     cfg_into_iter!(0..num_constraints)
         .map(|constraint| {
-            (0..=max_degree)
+            // Skip building coefficient MLEs for zero-ideal constraints.
+            // For an honest prover these MLEs are zero; the combined
+            // polynomial resolver handles the zero entries downstream.
+            if skip_constraints[constraint] {
+                return vec![];
+            }
+            cfg_into_iter!(0..=max_degree)
```
amit0365: We should keep this sequential since max_degree of constraints is usually small < 10

piop/src/combined_poly_resolver.rs
```diff
@@ -178,7 +177,8 @@ impl<F: InnerTransparentField + FromPrimitiveWithConfig + Send + Sync> CombinedP
         let evals: Vec<F> = sumcheck_prover_state.mles[2..]
             .iter()
             .map(|mle| {
-                mle.evaluate_with_config(slice::from_ref(last_sumcheck_challenge), field_cfg)
+                mle.clone()
```
amit0365: `sumcheck_prover_state.mles` aren't used after this point so we could use drain() to avoid cloning.

piop/src/ideal_check/combined_poly_builder.rs
```diff
@@ -129,28 +134,151 @@ fn prepare_coefficient_mles<F: PrimeField>(
     num_constraints: usize,
     max_degree: usize,
     max_degrees_and_combined_poly_rows: &[(usize, Vec<DynamicPolynomialF<F>>)],
-    field_zero: &F,
+    field_zero_inner: &F::Inner,
```
amit0365: We should probably change naming to `zero_as_field_inner` for clarity.

piop/src/projections.rs
```diff
@@ -33,13 +47,119 @@ where
...
+        .collect();
+
+    // Build row-by-row
+    cfg_iter_mut!(result)
+        .enumerate()
+        .for_each(|(row_idx, row)| {
+            let spare = row.spare_capacity_mut();
+
+            // Binary poly columns
+            cfg_iter_mut!(spare[..binary_len])
```
amit0365, line 75: Currently we have 3 rayon dispatches per row. We should merge these into a single inner parallel pass over all columns (e.g. over spare[..num_cols] with index-based approach).

piop/src/projections.rs
```diff
@@ -33,13 +47,119 @@ where
...
+                                } else {
+                                    zero.clone()
+                                }
+                            })
+                            .collect(),
+                    );
+                });
+
+            // Arbitrary poly columns
+            cfg_iter_mut!(spare[binary_len..binary_len + arbitrary_len])
```
amit0365, line 94: ditto

piop/src/projections.rs
```diff
@@ -33,13 +47,119 @@ where
...
+                    slot.write(
+                        arbitrary_poly
+                            .iter()
+                            .map(|coeff| F::from_with_cfg(coeff.clone(), field_cfg))
+                            .collect(),
+                    );
+                });
+
+            // Int columns
+            cfg_iter_mut!(spare[binary_len + arbitrary_len..])
```
amit0365, line 107: ditto

uair/src/degree_counter.rs
amit0365, line 161: Should this be `mul_assign`? Unless I am missing something. @ElijahVlasov

amit0365 (changes requested):

uair/src/degree_counter.rs
frozenspider: This "semiring" is for tracking max degree only, so addition/subtraction is max of two max degrees, and multiplication is sum of max degrees

piop/src/ideal_check/combined_poly_builder.rs
frozenspider: It makes no difference in current test UAIRs, but I guess on a real UAIR that might hurt the performance indeed

piop/src/projections.rs
frozenspider: Tried that - it doesn't lead to any performance change (in either side), but the code becomes way less readable, so the effect is net negative

piop/benches/ideal_check.rs
```diff
@@ -63,7 +69,9 @@ fn bench_no_mult<const INT_LIMBS: usize, const FIELD_LIMBS: usize>(
                     .collect()
             });
 
-        IdealCheckProtocol::prove_as_subprotocol::<TestAirNoMultiplication<INT_LIMBS>>(
+        // Even though this UAIR is linear, using prove_combined yields much better
```
amit0365, line 72: How can we account for this?

piop/benches/ideal_check.rs
amit0365, line 1: We should consider adding prove_combined bench for BigLinearUair to compare performance with MLE-first approach.

piop/benches/ideal_check.rs
frozenspider: There is no trivial heuristic, we could try calculating number of operations per row, but that's expensive and an overkill - based on discussion [here](https://demerzelsolutions.slack.com/archives/C08S7BPF50B/p1771590650306469?thread_ts=1771330611.110359&cid=C08S7BPF50B), some manual configuration is okay we'll probably have more as we add more UAIRs

piop/benches/ideal_check.rs
frozenspider: What would be the point of it though? `BigLinearUair` is our toy UAIR, I think our focus would be better placed on implementing real-world UAIRs and playing with them instead

amit0365: It would give us concrete performance gains for MLE first approach. But it seems from this slack [thread](https://demerzelsolutions.slack.com/archives/C08S7BPF50B/p1771590650306469?thread_ts=1771330611.110359&cid=C08S7BPF50B) we dont get improvements for BigLinearUair

frozenspider: We do get improvements from using MLE-first approach - see PR description for the approximate numbers.
Adding a separate BigLinearUair using an old approach to the bench suite doesn't sound productive though - it will noticeably increase bench suite running time, but the numbers returned wont be of any use to us.

amit0365: The PR description conflates speedups from the MLE-first approach with speedups from other sources. For example, the 2x speedup in NoMult comes from optimizations other than MLE-first. A separate bench for BigLinearUair using an old approach would be useful to isolate the signal.

frozenspider: You might have a point. Added both benches.

amit0365, line 1: LGTM! Could you update the numbers for BigLinearUair to compare both approaches

amit0365 (changes requested):

piop/benches/ideal_check.rs
frozenspider: Added

amit0365 (approved):

PR #132, opened 2026-02-26
Reworked Zip+ to allow flexible transcript usage

frozenspider:
* Singular `PcsTranscript` has been split onto `PcsProverTranscript` and `PcsVerifierTranscript`. Both transcripts are initialized with commitment hash. Once we support public inputs, they should be included too, but this is out of scope of Zip+.
* Instead of using type safety wrapper `ZipPlusProof`, we're now passing `&mut transcript` through all phases. This reduces protection against misuse, but allows using the same transcript for multiple Zip+ runs, and for other subprotocols.
* Added `prove_single` helper method.

This PR does not affect benchmarks.


frozenspider:
@codex review

maksimryndin:
@codex review

zip-plus/src/pcs/phase_verify.rs
```diff
@@ -336,51 +350,80 @@ mod tests {
...
     #[test]
     fn verification_fails_with_tampered_proof() {
-        fn tamper(proof: ZipPlusProof) -> ZipPlusProof {
-            let mut tampered = proof.0.clone();
+        fn tamper(mut proof: PcsVerifierTranscript) -> PcsVerifierTranscript {
             // Byte 0 is the 1-byte LENGTH_NUM_BYTES prefix for b field elements.
             // Flip byte 1 (first byte of the first b element's VALUE) instead.
-            tampered[1] ^= 0x01;
-            ZipPlusProof(tampered)
+            proof.stream.get_mut()[0] ^= 0x01;
```
amit0365: We should keep this to index 1. It corrupts the field element to test soundness.

zip-plus/src/pcs/phase_verify.rs
```diff
@@ -828,28 +1116,48 @@ mod tests {
...
+                &comm,
+                &field_cfg,
+                &point_f,
+                &eval_f,
+            )
+            .expect("verify");
         }
 
-        inner::<19>();
+        inner::<12>();
```
amit0365, line 1169: What was the reason for this change?

zip-plus/src/pcs_transcript.rs
```diff
@@ -139,10 +128,10 @@ impl PcsTranscript {
         F::Modulus: Transcribable,
     {
         self.fs_transcript.absorb_random_field(fe, buf);
-        fe.inner().write_transcription_bytes(buf);
-        self.stream.write_all(buf)?;
         fe.modulus().write_transcription_bytes(buf);
         self.stream.write_all(buf)?;
+        fe.inner().write_transcription_bytes(buf);
```
amit0365, line 133: Writing the modulus before the field element makes sense. Was there any other reason?

amit0365 (changes requested):

zip-plus/src/pcs/phase_verify.rs
frozenspider: `<19>` was committed by mistake in one of the previous PRs, I tweaked that value to test a specific case, it's supposed to be `<12>`.
The reason not to keep it at 19 is that this test is exceedingly slow, taking 15 seconds on my machine.

zip-plus/src/pcs_transcript.rs
frozenspider: Not really, just feels more logical to do it this way

zip-plus/src/pcs/phase_verify.rs
frozenspider: Good catch! Although index should no longer be 1

frozenspider: Fixed, and added sanity check that we're doing the tampering correctly

amit0365 (approved):

PR #130, opened 2026-02-23
Update Zip+ PCS 

amit0365:
This PR addresses https://github.com/NethermindEth/zinc-plus/issues/104 and https://github.com/NethermindEth/zinc-plus/issues/114

PR #133, opened 2026-02-27
transcript: switch Fiat-Shamir hashing from Keccak-256 to BLAKE3

albert-garreta:
## Summary

Replace the Keccak-256 (sha3) based Fiat-Shamir transcript with BLAKE3.

## Changes

- **transcript/Cargo.toml**: Replace `sha3 = "0.10"` with `blake3 = "1"`. The `asm` feature is now a no-op since BLAKE3 auto-detects SIMD at runtime.
- **transcript/src/lib.rs**: New `Blake3Transcript` struct:
  - Uses BLAKE3's incremental `Hasher` -- multiple `update()` calls are treated as one contiguous message internally.
  - Squeezing uses `clone() + finalize_xof()` (XOF mode) for arbitrary-length output, replacing the old counter-based loop that cloned the Keccak hasher repeatedly.
  - `absorb_random_field()` serialises separator bytes + modulus + inner value into one contiguous buffer and calls `update()` once (was 7 tiny calls per element).
  - `absorb_random_field_slice()` serialises the entire slice into a single buffer before one `update()` call, enabling BLAKE3's SIMD tree hashing on large chunks.
  - `KeccakTranscript` kept as a type alias for backward compatibility -- no changes needed in downstream crates.

## Motivation

When profiling LogUp, transcript absorption of large field-element slices was a significant cost. The old Keccak implementation made 7 tiny `hasher.update()` calls per field element, preventing the hash function from leveraging SIMD. Batching into contiguous buffers and using BLAKE3 (which has native SIMD tree hashing) addresses this.

## Testing

- All 33 `zinc-piop` tests pass
- All 35 `zinc-poly` tests pass
- Full workspace compiles cleanly


frozenspider:
Let's wait for #132 and #130 before adopting this (only the PCS part though)

frozenspider:
Closed in favor of #163

PR #135, opened 2026-03-03
Accommodate crypto-primitives change

frozenspider:
This is an update to accommodate changes made in https://github.com/NethermindEth/crypto-primitives/pull/31:
* `Field::Modulus` is now separate from `Field::Inner`
* Switched to `into_inner()` to avoid `inner().clone()` where it's trivial to do

### DIFF
--- a/transcript/src/lib.rs
+++ b/transcript/src/lib.rs
@@ -92,11 +92,19 @@ impl Transcript for KeccakTranscript {
     /// Absorbs a field element into the transcript.
     /// Delegates to the field element's implementation of
     /// absorb_into_transcript.
+    // Note: Currently this only works for fields whose modulus and inner element
+    // have the same byte length
     fn absorb_random_field<F>(&mut self, v: &F, buf: &mut [u8])
     where
         F: PrimeField,
         F::Inner: Transcribable,
+        F::Modulus: Transcribable,
     {
+        debug_assert_eq!(F::Inner::LENGTH_NUM_BYTES, F::Modulus::LENGTH_NUM_BYTES);
+        debug_assert_eq!(
+            F::Inner::get_num_bytes(v.inner()),
+            F::Modulus::get_num_bytes(&v.modulus())
+        );
         self.absorb(&[0x3]);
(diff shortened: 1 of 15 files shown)


frozenspider:
Until this merges, builds will continue to be broken (including coverage PR check on this one, as it cannot build master)

transcript/src/lib.rs
```diff
@@ -92,11 +92,19 @@ impl Transcript for KeccakTranscript {
     /// Absorbs a field element into the transcript.
     /// Delegates to the field element's implementation of
     /// absorb_into_transcript.
+    // Note: Currently this only works for fields whose modulus and inner element
+    // have the same byte length
```
amit0365, line 96: Could you explain why in the comments

zip-plus/src/pcs_transcript.rs
```diff
@@ -52,14 +52,20 @@ impl PcsTranscript {
         }
     }
 
+    // Note: Currently this only works for fields whose modulus and inner element
+    // have the same byte length
```
amit0365, line 56: ditto

zip-plus/src/pcs_transcript.rs
```diff
@@ -76,15 +82,20 @@ impl PcsTranscript {
         Ok(())
     }
 
+    // Note: Currently this only works for fields whose modulus and inner element
+    // have the same byte length
```
amit0365, line 86: ditto

transcript/src/lib.rs
frozenspider: This is simply because that's the only use case we used to have, and we need these changes ASAP to make builds and commit checks work.
I doubt we'll ever encounter a case where this isn't true.

amit0365: This would break for extension fields though. But we don't use them so it's fine.

frozenspider: We do not use extension fields, nor do we plan to use them, to my knowledge. And if we decide to do so, there would be a lot more to be tweaked, not just this.

amit0365 (approved):

PR #136, opened 2026-03-03
End-to-end Zinc+ PIOP protocol (without proving down evals)

frozenspider:
(Depends on #132)
First step toward end-to-end Zinc+ PIOP protocol.
Down evaluation proving is missing and is done via `resolve_subclaim` placeholder for now.

Additional changes:
* Evaluation phase of Zip+ split to allow passing points already mapped to F
* UAIRs made generic over semirings
* Generalized `FromRef<Boolean>` for all `From<bool>`
* Removed `WideningMulByScalar`, instead tweaked `MulByScalar` to widen the result when needed.
  * As such, IPRS codes no longer require parameterizing by `WideningMulByScalar`.
  * `MulByTwiddle` will also be removed in a separate PR.
* In `MBSInnerProduct`, replaced `mapped_inner_product` with `inner_product_field`.


agarreta:
*Albert* : I observed significant gains from doing this in another branch


*Ai agent*

Suggestion: parallelize Step 0 (PCS commits)

The commit phase iterates over witness columns sequentially:

for col in witness {
    let (hint, comm) = ZipPlus::<Zt, Lc>::commit(pp, col)?;
    hints.push(hint);
    commitments.push(comm);
}

Since each column's commitment is independent, this could use rayon::par_iter (or similar) to commit all columns in parallel. The three column-type batches (bin, arb, int) are also independent of each other and could be committed concurrently. This should give a meaningful speedup for traces with many columns or large polynomial sizes.

agarreta:
Do we have an MLE-first path for the ideal check when the degree is 1? It is quite important performance wise

frozenspider:
> _Albert_ : I observed significant gains from doing this in another branch
> 
> _Ai agent_
> 
> Suggestion: parallelize Step 0 (PCS commits)
> 
> The commit phase iterates over witness columns sequentially:
> 
> for col in witness { let (hint, comm) = ZipPlus::<Zt, Lc>::commit(pp, col)?; hints.push(hint); commitments.push(comm); }
> 
> Since each column's commitment is independent, this could use rayon::par_iter (or similar) to commit all columns in parallel. The three column-type batches (bin, arb, int) are also independent of each other and could be committed concurrently. This should give a meaningful speedup for traces with many columns or large polynomial sizes.

Once we finalize batched Zip+, this might not be relevant anymore I think.

agarreta:
Zip+ is not using the batched version. That's essential for performance

frozenspider:
> Do we have an MLE-first path for the ideal check when the degree is 1? It is quite important performance wise

Nope, that's a separate PR: https://github.com/NethermindEth/zinc-plus/pull/128

frozenspider:
@agarreta, once batched Zip+ and MLE-first PRs land, we'll get those bonuses for free. What's your take on the correctness part?

maksimryndin:
@codex review

poly/src/utils.rs
amit0365, line 249: We should acknowledge the source in comments. This is copied from - https://github.com/TomWambsgans/Whirlaway/blob/9e3592b5b5c3702fe3a1728be758506a7ee9283b/crates/air/src/utils.rs#L92

protocol/src/verifier.rs
```diff
@@ -0,0 +1,215 @@
...
+        DenseMultilinearExtension<F::Inner>: MultilinearExtensionWithConfig<F>,
+    {
+        let num_cols = projected_trace_f.len();
+
+        // Check "down" evaluations (shifted/next-row columns).
+        // The shifted trace drops the first row and zero-pads, matching
+        // the CombinedPolyResolver's convention.
+        for (i, (mle, expected)) in projected_trace_f
+            .iter()
+            .zip(subclaim.down_evals.iter())
```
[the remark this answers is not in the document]
frozenspider: This is pretty much a temporary placeholder so it doesn't matter

utils/src/inner_product.rs
amit0365: This could overflow for other types. I missed this when adding it. Could you please add a comment - Cannot overflow because we only call this with Field type

protocol/src/lib.rs
```diff
@@ -0,0 +1,611 @@
...
+/// Proof produced by the Zinc+ PIOP for UCS.
+///
+/// Contains the two subproofs from Steps 2 and 4:
+/// - `ideal_check`: MLE evaluations in F_q\[X\] (Step 2).
+/// - `resolver`: sumcheck proof + trace evaluation claims (Step 4).
+#[derive(Clone, Debug)]
+pub struct Proof<F: PrimeField> {
+    /// Number of witness columns of each type (binary polynomials, arbitrary
+    /// polynomials, integers).
+    pub num_witness_cols: (usize, usize, usize),
```
amit0365: This is redundant as the verifier can derive this from respective commitment batch size instead. A malicious proof with wrong num_cols will be eventually rejected by the verifier but return a vague `InvalidPcsOpen` error

protocol/src/verifier.rs
```diff
@@ -0,0 +1,261 @@
...
+        // Absorb lifted_evals, then check \psi_a consistency:
+        // \psi_a(lifted_eval_j) must equal open_eval_j.
+        let mut transcription_buf: Vec<u8> = vec![0; F::Inner::NUM_BYTES];
+        for bar_u in &proof.lifted_evals {
+            pcs_transcript
+                .fs_transcript
+                .absorb_random_field_slice(&bar_u.coeffs, &mut transcription_buf);
+        }
+
+        for (j, (bar_u, up_eval)) in proof
```
amit0365, line 141: We should use assert_eq here to catch malformed values instead of returning vague `InvalidPcsOpen` error

protocol/src/lib.rs
```diff
@@ -0,0 +1,611 @@
...
+    #[error("PCS verification failed at column {0}: {1}")]
+    PcsVerification(usize, ZipError),
+}
+
+//
+// Tests
+//
+
+#[cfg(test)]
+mod tests {
```
amit0365, line 207: We should add negative tests too

protocol/src/prover.rs
```diff
@@ -0,0 +1,272 @@
...
+    ///    prime q, project trace and scalars.
+    /// 2. **Ideal check**: sample `r in F_q^mu`, prover sends MLE evaluations,
+    ///    verifier checks ideal membership.
+    /// 3. **Evaluation projection** (`\psi_a`: `F_q[X] -> F_q`): sample `a in
+    ///    F_q`, evaluate polynomials at `X = a`.
+    /// 4. **F_q sumcheck**: finite-field sumcheck proving the projected
+    ///    constraint claim. Produces `up_evals` and `down_evals` at point `r'`.
+    /// 5. **Lift-and-project**: compute per-column polynomial MLE evaluations
+    ///    at `r'` (in `F_q[X]`, before `\psi_a`). Absorb into transcript.
+    /// 7. **PCS open**: Zip+ prove for each committed column, proving witness
```
amit0365: Minor typo should be 6

protocol/src/lib.rs
```diff
@@ -0,0 +1,611 @@
...
+
+        type BinaryLc = RaaCode<Self::BinaryZt, TestRaaConfig, RAA_REP>;
+        type ArbitraryLc = RaaCode<Self::ArbitraryZt, TestRaaConfig, RAA_REP>;
+        type IntLc = RaaCode<Self::IntZt, TestRaaConfig, RAA_REP>;
+    }
+
+    /// Helper: project a DensePolynomial scalar to DynamicPolynomialF
+    /// by projecting each coefficient via φ_q.
+    fn project_scalar_fn<R>(
+        scalar: &DensePolynomial<R, 32>,
```
amit0365, line 385: Can use defined `DEGREE_PLUS_ONE` instead of hardcoding

poly/src/univariate/binary_u64.rs
amit0365: References deleted `BinaryRefPolyWideningMulByScalar` for trivial comparison

zip-plus/src/pcs_transcript.rs
```diff
@@ -65,8 +65,24 @@ impl PcsProverTranscript {
         Ok(result)
     }
 
-    pub fn reserve_capacity(&mut self, capacity: usize) {
-        self.stream.get_mut().reserve(capacity)
+    pub fn new_from_commitments<'a>(
+        comms: impl Iterator<Item = &'a ZipPlusCommitment>,
+    ) -> Result<Self, ZipError> {
```
amit0365: We can just return Self here as there is no fallible operation here

zip-plus/src/pcs_transcript.rs
amit0365: Minor pre-existing typo - again it into

amit0365 (changes requested):

protocol/src/lib.rs
frozenspider: While that is true at this stage, they will in fact be used in a follow-up PR #142 so I suggest we keep them

protocol/src/lib.rs
frozenspider: This is merely a test function. However, in #140 it's also used in benches so it was rewritten to be fully generic. Again, my suggestion is to keep it the way it is for this specific PR.

poly/src/univariate/binary_u64.rs
frozenspider: Good catch! Even worse, this specific check tests `result_ref.coeffs` being equal to itself, clearly a mistake.

utils/src/inner_product.rs
frozenspider: We should probably make this "mapped" version field-specific instead, what do you think?

protocol/src/verifier.rs
frozenspider: While this check is a mere sanity check, it seems fitting to return a proper error here (also it's lifted evaluation error, not PCS error).
However, this whole logic is reworked in #140 because of multipoint evaluation, so I would rather not fix it here.

protocol/src/lib.rs
amit0365: I looked at #142 and don't see a reason why the verifier can't derive this from the commitments. Could you elaborate? The proof struct should be minimal to reduce attack surface

utils/src/inner_product.rs
amit0365: Yes thats a better approach. Feel free to make this field specific. Otherwise leave a comment so we don't forget it. I can fix this in my next PR.

protocol/src/lib.rs
frozenspider: You're right, my bad - we can always get those values from UAIR signature

utils/src/inner_product.rs
frozenspider: Done: https://github.com/NethermindEth/zinc-plus/pull/136/commits/04c97059f5a3388567a2d3706b39015a12fdd675

protocol/src/lib.rs
```diff
@@ -608,4 +605,70 @@ mod tests {
         )
         .expect("Subclaim resolution failed");
     }
+
+    //
+    // Negative tests for BigLinearUair: verify that proof tampering is detected.
+    //
+
+    fn big_linear_verify_tampered(tamper: impl FnOnce(&mut Proof<F>)) {
```
amit0365, line 613: Cool helper function!

amit0365 (approved):

PR #140, opened 2026-03-11
End-to-end Zinc+ PIOP protocol: Down evaluations

### LINKED ISSUE #92: Shifted trace evaluations resolver
After combined poly resolver stage we end up with evaluations of `up` MLEs, a.k.a. the trace, and evaluations of `down` MLEs, a.k.a. shifted trace. While the former can be further discharged by ZIP+ the latter is not that straightforward. 

In Whirlaway they define the following expression:

Image
but our definition of `c^down` is even simpler, we don't clamp values at the end of the trace but put zero at the last row.
So, the expression becomes 

Image
where `next` MLE is the MLE that's defined on hypercube as the characteristic function of the graph of `f(x)=x+1`. For more information see section 5.1 of https://eprint.iacr.org/2023/552.pdf:

Image

This `next` MLE and an algorithm of log-time computing `next` evaluations on the side of the verifier is defined in https://github.com/NethermindEth/zinc-plus/pull/96.

From here one needs to instantiate the sumcheck protocol to handle a degree-2 polynomial over MLEs `next` and `c^up` (can be named `up` or `trace` in different parts of the code). Leading to evaluation claims about `next` and `c^up`. The former verifier discharges themselves using the aforemention log-time algorithm, the latter is passed further to be later discharched by ZIP+.

frozenspider:
(Also resolves #92)

Extend #136 with the proof of `down` evaluations. They are proven through their `up` counterparts, by running an additional subprotocol we call "multipoint evaluation" that reduces `up/down` evaluation claims at one point to `up` evaluation claim at a different point. This allows us to prove both `up` and `down` evals with a single set of Zip+.

This PR also introduces end-to-end benchmarks using our linear UAIRs.

Note that we're still missing **public parameters** (#142) and **lookup argument**!


frozenspider:
@codex review

protocol/src/prover.rs
```diff
@@ -254,19 +225,72 @@ where
...
+fn compute_lifted_evals<F, const D: usize>(
+    point: &[F],
+    trace_bin_poly: &[DenseMultilinearExtension<BinaryPoly<D>>],
+    projected_trace: &[DenseMultilinearExtension<DynamicPolynomialF<F>>],
+    field_cfg: &F::Config,
+) -> Vec<DynamicPolynomialF<F>>
+where
+    F: InnerTransparentField,
+{
+    let eq_table = zinc_poly::utils::build_eq_x_r_vec(point, field_cfg)
```
amit0365, line 260: Should we use `?` instead of panicking? This way the error can propagate to the prover.

protocol/src/prover.rs
```diff
@@ -254,19 +225,72 @@ where
...
+        .expect("compute_lifted_evals: eq table build failed");
+
+    let n_bin = trace_bin_poly.len();
+    let zero = F::zero_with_cfg(field_cfg);
+
+    let mut result = Vec::with_capacity(projected_trace.len());
+
+    for col in trace_bin_poly {
```
amit0365: This can be parallelized

protocol/src/prover.rs
```diff
@@ -254,19 +225,72 @@ where
...
+            for (l, coeff) in entry.iter().enumerate() {
+                if coeff.into_inner() {
+                    coeffs[l] += &eq_table[b];
+                }
+            }
+        }
+        result.push(DynamicPolynomialF::new_trimmed(coeffs));
+    }
+
+    for col_mle in &projected_trace[n_bin..] {
```
amit0365: ditto

protocol/benches/e2e.rs
```diff
@@ -0,0 +1,473 @@
...
+        + for<'a> MulByScalar<&'a F>
+        + FromRef<F>
+        + Send
+        + Sync
+        + 'static,
+    F: for<'a> FromWithConfig<&'a Zt::Int>,
+    <F as Field>::Modulus: ConstTranscribable + FromRef<Zt::Fmod>,
+    U: Uair + 'static,
+    IdealOverF: Ideal + IdealCheck<DynamicPolynomialF<F>>,
+    U: Uair + 'static,
```
amit0365: Duplicate bound

amit0365 (changes requested):

protocol/src/prover.rs
frozenspider: Hmm. In this specific case, the error might only happed due to a bug in a prover (zero-length point or logic bug), so I would say panicking is in order

protocol/src/prover.rs
frozenspider: Improves parallelized running time by a ~5% for BigLinearUair

amit0365 (approved):

PR #142, opened 2026-03-12
End-to-end Zinc+ PIOP protocol: Public input columns

### LINKED ISSUE #108: Public inputs in UAIRs and subprotocols

frozenspider:
Resolves #108

* UAIR signature now specifies how many columns are public (they always come first in every category).
* PIOP prover new sends lifted evaluations only for witness columns. Verifier computes them for public columns on its own.
* Mutipoint evaluation subprotocol verification has been tweaked to work in two stages, to return `r0` point before finally validating open evaluations. This is used by PIOP verifier to reconstruct open evaluations at `r0`.
* Added a variant of `BigLinearUair` that has first 4 binary columns marked as public. It's added to tests and benchmark as well.
* Additionally, fixed an issue with binary polynomial being incorrectly transcribed, added a regression test for it.

### DIFF
--- a/protocol/src/verifier.rs
+++ b/protocol/src/verifier.rs
@@ -6,9 +6,12 @@ use zinc_piop::{
     combined_poly_resolver::CombinedPolyResolver,
     ideal_check::IdealCheckProtocol,
     multipoint_eval::MultipointEval,
-    projections::{project_scalars, project_scalars_to_field},
+    projections::{project_scalars, project_scalars_to_field, project_trace_coeffs_row_major},
 };
-use zinc_poly::{EvaluatablePolynomial, univariate::dynamic::over_field::DynamicPolynomialF};
+use zinc_poly::{
+    EvaluatablePolynomial, mle::DenseMultilinearExtension,
+    univariate::dynamic::over_field::DynamicPolynomialF,
+};
 use zinc_transcript::{
     KeccakTranscript,
     traits::{ConstTranscribable, Transcript},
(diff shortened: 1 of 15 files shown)


protocol/src/verifier.rs
```diff
@@ -128,39 +140,73 @@ where
...
+        let public_lifted = if add!(add!(num_pub_bin, num_pub_arb), num_pub_int) > 0 {
+            let projected_public = project_trace_coeffs_row_major::<F, Zt::Int, Zt::Int, D>(
+                public_bin, public_arb, public_int, &field_cfg,
+            );
+            compute_lifted_evals::<F, D>(r_0, public_bin, &projected_public, &field_cfg)
+        } else {
+            Vec::new()
+        };
+
+        let all_lifted_evals: Vec<_> = public_lifted[..num_pub_bin]
```
amit0365, line 181: Is interleaving public cols necessary for the protocol? Switching to all public cols first then witness cols would make it cleaner and less error-prone.

amit0365 (approved):

protocol/src/verifier.rs
frozenspider: It's not necessary, no, just seems more logical to structure it this way.
I agree that this approach is error-prone, and it has been partially addressed in [the follow-up PR](https://github.com/NethermindEth/zinc-plus/pull/148) by introducing `UairTrace` struct.
(It won't help in this specific place though, as we're working with lifted evals and not raw trace)

I'll give a "all-public-then-all-private" approach a try, and if that structure would look better, make a separate follow-up PR.

PR #143, opened 2026-03-13
Generalize shifts to arbitrary per-column amounts

amit0365:
## Summary

- Add `ShiftSpec`, `ColumnLayout` on `UairSignature` for per-column shift declarations
- Sparse down-trace in CPR and ideal check via `shifts()` / `down_layout()`
- Add `eval_shift_predicate` for shift-by-c MLE evaluation
- Add `build_next_c_r_mle` for shift-by-c selector MLE construction
- Generalize multipoint_eval from shift-by-1 to arbitrary per-column shift amounts with separate up/down challenge batching
- Add `TestUairMixedShifts` with shift-by-1 and shift-by-2
- End-to-end protocol test with mixed shifts through full pipeline
-  Improve `next_mle_eval` from O(n²) to O(n)
- Tightened negative tests in E2E to assert specific error variants, ensuring each tamper is caught at the expected protocol step

## Context

Makes the down-trace sparse: each UAIR declares which columns are shifted and by how much via `ShiftSpec`. The multipoint_eval now proves: `Σ_j γ_j · up_j + Σ_k α_k · down_k` with one `next_c` kernel per shifted column (ungrouped). A follow-up can group by shift amount to reduce MLE count from `2+2K` to `2(s+1)`.

## Test plan

- [x] `cargo test -p zinc-poly` — `build_next_c_r_mle` + `next_mle_eval` tests
- [x] `cargo test -p zinc-piop -- multipoint_eval` — 16 unit tests (happy path + negative, mixed shifts)
- [x] `cargo test -p zinc-piop -- shift_predicate` — predicate tests
- [x] `cargo test -p zinc-test-uair` — all test UAIRs including `TestUairMixedShifts`
- [x] `cargo test -p zinc-protocol` — 10 e2e tests including `test_e2e_mixed_shifts`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/piop/src/ideal_check/combined_poly_builder.rs
+++ b/piop/src/ideal_check/combined_poly_builder.rs
@@ -12,7 +12,7 @@ use zinc_poly::{
     univariate::dynamic::over_field::DynamicPolynomialF,
 };
 use zinc_uair::{
-    ConstraintBuilder, TraceRow, Uair,
+    ColumnLayout, ConstraintBuilder, TraceRow, Uair,
     degree_counter::{count_constraint_degrees, count_max_degree},
     ideal::ImpossibleIdeal,
 };
@@ -39,6 +39,8 @@ where
     U: Uair,
 {
     let field_zero = F::zero_with_cfg(field_cfg);
+    let uair_sig = U::signature();
+    let down_layout = uair_sig.down_cols().as_column_layout();
 
     let num_rows = trace_matrix.len();
 
@@ -46,13 +48,25 @@ where
         cfg_into_iter!(0..num_rows - 1)
             .map(|row_idx| {
                 let up = &trace_matrix[row_idx];
-                let down = &trace_matrix[row_idx + 1];
+
+                let down: Vec<DynamicPolynomialF<F>> = uair_sig
+                    .shifts()
+                    .iter()
+                    .map(|spec| {
+                        if row_idx + spec.shift_amount() < num_rows {
+                            trace_matrix[row_idx + spec.shift_amount()][spec.source_col()].clone()
+                        } else {
+                            DynamicPolynomialF::new([]) // zero padding
+                        }
+                    })
+                    .collect();
 
                 combine_rows_and_get_max_degree::<F, U>(
                     up,
-                    down,
+                    &down,
                     num_constraints,
                     projected_scalars,
+                    down_layout,
                 )
             })
             .collect();
@@ -92,12 +106,12 @@ fn combine_rows_and_get_max_degree<F, U>(
     down: &[DynamicPolynomialF<F>],
     num_constraints: usize,
     projected_scalars: &HashMap<U::Scalar, DynamicPolynomialF<F>>,
+    down_layout: &ColumnLayout,
 ) -> (usize, Vec<DynamicPolynomialF<F>>)
 where
     F: PrimeField,
     U: Uair,
 {
-    let sig = U::signature();
     let mut constraint_builder = CombinedPolyRowBuilder::new(num_constraints);
 
     let project = |x: &U::Scalar| {
@@ -109,8 +123,8 @@ where
 
     U::constrain_general(
         &mut constraint_builder,
-        TraceRow::from_slice_with_signature(up, &sig),
-        TraceRow::from_slice_with_signature(down, &sig),
+        TraceRow::from_slice_with_layout(up, U::signature().total_cols().as_column_layout()),
+        TraceRow::from_slice_with_layout(down, down_layout),
         &project,
         |x, y| Some(project(y) * x),
         ImpossibleIdeal::from_ref,
@@ -175,7 +189,8 @@ fn prepare_coefficient_mles<F: PrimeField>(
 ///
 /// `trace_matrix` is column-indexed: `trace_matrix[col]` is an MLE.
 ///
-/// Does `2 * num_columns * max_num_coeffs` evaluations of MLEs.
+/// Does `(num_columns + num_shifted_columns) * max_num_coeffs` evaluations of
+/// MLEs.
 #[allow(clippy::arithmetic_side_effects)]
 pub fn evaluate_combined_polynomials<F, U>(
     trace_matrix: &ColumnMajorTrace<F>,
@@ -192,7 +207,6 @@ where
     let zero_inner = field_zero.inner().clone();
     let num_rows = trace_matrix.first().map(|c| c.len()).unwrap_or(0);
     let num_vars = evaluation_point.len();
-    let sig = U::signature();
 
     // Sanity check: this approach only works for linear constraints
     if count_max_degree::<U>() > 1 {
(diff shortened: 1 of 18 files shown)


frozenspider:
@codex review

amit0365:
@codex review

amit0365:
@codex review

frozenspider:
@codex review

uair/src/lib.rs
```diff
@@ -37,68 +37,254 @@ pub trait ConstraintBuilder {
...
+        self.binary_poly_cols
     }
 
-    /// Maximum number of columns across the three types (public + witness per
-    /// type).
+    pub fn arbitrary_poly_cols(&self) -> usize {
+        self.arbitrary_poly_cols
+    }
+
+    pub fn int_cols(&self) -> usize {
```
frozenspider: Naming, here and everywhere, is confusing - these aren't actual columns, just their number. This should be reflected in the name.

uair/src/lib.rs
```diff
@@ -37,68 +37,254 @@ pub trait ConstraintBuilder {
...
+
+    fn compute_down_layout(
+        total_cols: &TotalColumnLayout,
+        shifts: &[ShiftSpec],
+    ) -> VirtualColumnLayout {
+        let bp_end = total_cols.binary_poly_cols();
+        let ap_end = add!(bp_end, total_cols.arbitrary_poly_cols());
+        let mut bp = 0usize;
+        let mut ap = 0usize;
+        let mut ic = 0usize;
```
frozenspider: Cryptic variable names

uair/src/lib.rs
```diff
@@ -37,68 +37,254 @@ pub trait ConstraintBuilder {
...
+    pub fn dummy_rows<T: Clone>(&self, val: T) -> (Vec<T>, Vec<T>) {
+        let up_size = self.total_cols.cols();
+        let down_size = self.down_cols.cols();
+        (vec![val.clone(); up_size], vec![val; down_size])
+    }
+
+    /// Witness column counts (total minus public) per type.
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn witness_cols(&self) -> WitnessColumnLayout {
+        WitnessColumnLayout::new(
```
frozenspider: How about we just compute this in the constructor and store in a field?
Also it seems that we don't really want all these getters, just making fields public is enough

uair/src/lib.rs
```diff
@@ -110,27 +296,31 @@ pub struct UairTrace<'a, PolyCoeff: Clone, Int: Clone, const D: usize> {
 }
 
 impl<PolyCoeff: Clone, Int: Clone, const D: usize> UairTrace<'static, PolyCoeff, Int, D> {
-    /// Returns a sub-trace containing only public columns.
-    /// Returned trace is borrowed from the full trace.
+    /// Returns a layout containing only public columns.
```
frozenspider: Old comment was changed incorrectly, we do not return a layout (as it was defined here), we return an actual sub-trace.
(Same for `witness`)

test-uair/src/lib.rs
```diff
@@ -34,10 +35,9 @@ where
     type Scalar = DensePolynomial<R, 32>;
 
     fn signature() -> UairSignature {
-        UairSignature {
-            witness_arbitrary_poly_cols: 3,
-            ..Default::default()
-        }
+        let total = TotalColumnLayout::new(0, 3, 0);
+        let shifts = (0..3).map(|i| ShiftSpec::new(i, 1)).collect();
+        UairSignature::new(total, PublicColumnLayout::default(), shifts)
```
frozenspider, line 40: How about we store a signature inside Uair struct instead of creating it on-demand?

test-uair/src/lib.rs
```diff
@@ -142,10 +142,8 @@ where
     type Scalar = DensePolynomial<R, 32>;
 
     fn signature() -> UairSignature {
-        UairSignature {
-            witness_arbitrary_poly_cols: 3,
-            ..Default::default()
-        }
+        let total = TotalColumnLayout::new(0, 3, 0);
+        UairSignature::new(total, PublicColumnLayout::default(), vec![])
```
frozenspider, line 146: Does not having shifts here (and below) change behaviour compared to what it was when down columns were shifted by 1 always?

test-uair/src/lib.rs
```diff
@@ -490,6 +477,98 @@ where
...
+            a_col.push(a_col[i].clone() + &b_col[i]);
+        }
+
+        // c[i] = b[i+2], zero-padded for last 2 entries
+        let mut c_col: Vec<DynamicPolynomialFS<R>> = Vec::with_capacity(n);
+        for i in 0..n {
+            if i + 2 < n {
+                c_col.push(b_col[i + 2].clone());
+            } else {
+                c_col.push(DynamicPolynomialFS::new(vec![]));
```
frozenspider: Nit: `DynamicPolynomialFS::zero()`

poly/src/utils.rs
```diff
@@ -213,15 +215,18 @@ where
...
+    }
 
-    // next_tilde(r, 0) = 0
-    // next_tilde(r, b) = eq(r, b-1) for b > 0
-    let mut evaluations = Vec::with_capacity(n);
-    evaluations.push(zero_inner);
-    evaluations.extend_from_slice(&eq_r.evaluations[..sub!(n, 1)]);
+    // next_c_tilde(r, 0) = 0 for b < c
+    // next_c_tilde(r, b - c) = eq(r, b - c) for b >= c
+    let mut evaluations = vec![zero_inner; c];
```
frozenspider: This is (marginally) worse that using `Vec::with_capacity` for known capacity

protocol/src/lib.rs
```diff
@@ -609,6 +612,20 @@ mod tests {
         );
     }
 
+    /// End-to-end test: TestUairMixedShifts.
+    ///
+    /// Uses mixed shift amounts (col a: shift 1, col b: shift 2).
+    /// Constraints: a[i+1] = a[i] + b[i], c[i] = b[i+2].
+    #[test]
+    fn test_e2e_mixed_shifts() {
+        do_test::<TestZincTypesRaa, TestUairMixedShifts<ZtInt>>(
```
frozenspider: We should use IPRS when possible, and since we have no multiplication here, it shouldn't be a problem

piop/src/ideal_check/combined_poly_builder.rs
```diff
@@ -39,20 +39,34 @@ where
...
-                let down = &trace_matrix[row_idx + 1];
+
+                let down: Vec<DynamicPolynomialF<F>> = uair_sig
+                    .shifts()
+                    .iter()
+                    .map(|spec| {
+                        if row_idx + spec.shift_amount() < num_rows {
+                            trace_matrix[row_idx + spec.shift_amount()][spec.source_col()].clone()
+                        } else {
+                            DynamicPolynomialF::new([]) // zero padding
```
frozenspider: `DynamicPolynomialF::zero()`

piop/src/shift_predicate.rs
```diff
@@ -0,0 +1,401 @@
...
+//! Evaluates `S_c(x, y)` — the multilinear extension of the shift-by-c
+//! indicator — at arbitrary field points.
+
+use crypto_primitives::PrimeField;
+use zinc_poly::utils::next_mle_eval;
+
+/// Evaluate the shift predicate `S_c(x, y)` at arbitrary field points.
+///
+/// Uses the high/low decomposition:
+///   `S_c(x, y) = L_0(x_lo, y_lo) · eq*(x_hi, y_hi)
```
frozenspider: What are `eq*` and `next*`, is it used in place of a tilde in MLE notation?

frozenspider (changes requested):

test-uair/src/lib.rs
amit0365: Good suggestion. But the shifts are usually small for e.g both SHA and ECDSA have ~12 shifts each. Moreover, storing signature inside Uair will require the callers to carry Uair as input at 20+ sites. I don't think this will be worth it.

protocol/src/lib.rs
amit0365: Is there a specific reason we should use IPRS and avoid RAA here?

piop/src/shift_predicate.rs
amit0365: Tilde and star are different notations used by the paper in Appendix: C. To simplify renamed them to eq/next_mle throughout.

test-uair/src/lib.rs
frozenspider: Hmm, the alternative to carrying around either UAIR or UAIR signature is the current approach of creating signature in-place, but this operation becomes heavier.
I agree that it's probably still negligible so how about leaving a TODO instead?

protocol/src/lib.rs
frozenspider: It's just to keep test suite aligned more closely with how the real protocol is used e.g. in benches

uair/src/lib.rs
amit0365: Clean! Storing witness cols as a field now. Keeping fields private is intentional - shifts should be sorted and column counts should be consistent. This prevents external code from constructing or mutating the struct into invalid states thus mitigating bugs.

piop/src/shift_predicate.rs
frozenspider: Aren't there only A and B appendixes in the Zinc+ paper though?

uair/src/lib.rs
frozenspider: Note that the latter is not as big problem in Rust as it is in many other languages due to the fact that mutability has to be declared explicitly, but I'm OK with either approach - just needs us to have more boilerplate.

piop/src/shift_predicate.rs
amit0365: The paper was updated with the shift predicate

test-uair/src/lib.rs
amit0365: With no shifts, we don't sample any challenges (alphas) now. Moreover, both the next and down mles don't enter the sumcheck.

test-uair/src/lib.rs
amit0365: Added

uair/src/lib.rs
amit0365: Agree on the mutability point.

frozenspider (approved):

PR #122, opened 2026-02-16
feat: add rate-1/4 IPRS configurations for F65537

albert-garreta:
## Summary

Add six new rate-1/4 IPRS code configurations for the Fermat prime field F65537 = 2^16 + 1.

## New configurations

| Config struct | BASE_LEN | BASE_DIM | Row lengths (D=1 / D=2 / D=3) |
|---|---|---|---|
| `PnttConfigF2_16R4B1` | 1 | 4 | 8 / 64 / 512 |
| `PnttConfigF2_16R4B2` | 2 | 8 | 16 / 128 / 1024 |
| `PnttConfigF2_16R4B4` | 4 | 16 | 32 / 256 / 2048 |
| `PnttConfigF2_16R4B16` | 16 | 64 | 128 / 1024 / 8192 |
| `PnttConfigF2_16R4B32` | 32 | 128 | 256 / 2048 / — |
| `PnttConfigF2_16R4B64` | 64 | 256 | 512 / 4096 / — |

All use rate 1/4 (`BASE_DIM = 4 * BASE_LEN`) and share the same 8th-roots-of-unity twiddle factors as the existing `PnttConfigF2_16_1`.

## Changes
- `zip-plus/src/code/iprs/pntt/radix8/params.rs`: struct definitions, `Config` trait implementations, twiddle tests
- `zip-plus/src/code/iprs.rs`: re-exports

PR #154, opened 2026-03-19
IPRS configs with rate 1/4

frozenspider:
Supersedes #122

* Added the configs introduced by #122
* Renamed configs to `F{MODULUS}_{BASE-LEN}_{BASE-DIM}` for simplicity and consistency.
* Changed E2E and Zip+ benches to use `32/128` config instead of `32/64`


zip-plus/src/code/iprs/pntt/radix8/params.rs
amit0365, line 1: Just a minor issue - For `PnttConfigF65537_64_256<3>`, `OUTPUT_LEN = 131072`, which exceeds F65537's max NTT domain of `2^16 = 65536`. So `Radix2EvaluationDomain::new` would panic at runtime. An assert check inside the macro would be better.

amit0365 (approved):

zip-plus/src/code/iprs/pntt/radix8/params.rs
frozenspider: Good idea!

Issue #120, opened 2026-02-15
Customizable matrix sizes in Zip+ and support for IPRS codes for polynomials with i32 types


albert-garreta:
Zip+ should support custom matrix sizes, not just square matrixes
Here's a vibe coded and vibed reviewed commit that achieves this: https://github.com/NethermindEth/zinc-plus/commit/83f7114f49463e5eef2cb3a6aed20960d120c3fa

We also need Zip+ to support using IPRS when the input polynomial is filled with i32, and not BPoly<31>

Note that IPRS codes require a separate config per matrix shape. I'll accordingly provide the matrix shapes we are interested in in our first use-case (SHA+ECDSA) and the corresponding IPRS code configs (these are in my branches rn --I'll extract the configs soon from there)

albert-garreta:
IPRS configs here https://github.com/NethermindEth/zinc-plus/pull/122

Concrete matrix dimensions we plan to use in the sha+ecdsa example:
1 x 2^10 matrices with BPoly<31>'s
2 x 2^9 matrices with i32's

PR #157, opened 2026-03-24
Arbitrary matrix sizes for linear code

frozenspider:
Goes towards #120

Based on #116, this PR allow specifying linear code row length directly. For IPRS code specifically, the row length is predefined by the config, but this is made explicit now.

Summary of changes:
* Removed universal `LinearCode::new(poly_size)` method - instead, RAA code has its own `new(poly_size)` while IPRS is simply created via `default`.
* Linear codes now universally implement `Debug` and `Eq`.
* Tweaked Zip+ type widths validation formula to accommodate non-square matrices
* Changed benchmarks to take in linear code instances instead of creating them in-place.
  * RAA codes are created with square matrices as before, so behaviour is unchanged.
* Tweaked tests to use IPRS codes instead of RAA, with the same 32/128 config we use in benches.
  * This requires us to increase number of variables (we need at least 8, but we use 10 in most places for good measure in order to have multiple rows)
  * Also requires widening`PolyZt::Cw` from `i32` to `i64`
* Fixed IPRS codes being unable to handle DEPTH == 0
* Fixed leftover: Zip+ benchmark still used `UNCHECKED` code instead of parameter being configurable via feature.

### DIFF
--- a/utils/src/mul_by_scalar.rs
+++ b/utils/src/mul_by_scalar.rs
@@ -48,8 +48,8 @@ macro_rules! impl_mul_int_by_primitive_scalar {
             impl<const LIMBS: usize, const LIMBS2: usize> MulByScalar<&$t, Int<LIMBS2>> for Int<LIMBS> {
                 #[allow(clippy::arithmetic_side_effects)] // By design
                 fn mul_by_scalar<const CHECK: bool>(&self, rhs: &$t) -> Option<Int<LIMBS2>> {
-                    if LIMBS2 < LIMBS {
-                        return None; // Cannot multiply if the left operand has fewer limbs than the output
+                    const {
+                        assert!(LIMBS <= LIMBS2, "Cannot multiply if the left operand has more limbs than the output");
                     }
                     let rhs: Int<LIMBS2> = Int::from_ref(rhs);
                     if CHECK {
--- a/utils/src/mul_by_scalar.rs
+++ b/utils/src/mul_by_scalar.rs
@@ -45,14 +45,17 @@ impl<const LIMBS: usize, const LIMBS2: usize> MulByScalar<&Int<LIMBS2>> for Int<
 macro_rules! impl_mul_int_by_primitive_scalar {
     ($($t:ty),*) => {
         $(
-            impl<const LIMBS: usize> MulByScalar<&$t> for Int<LIMBS> {
+            impl<const LIMBS: usize, const LIMBS2: usize> MulByScalar<&$t, Int<LIMBS2>> for Int<LIMBS> {
                 #[allow(clippy::arithmetic_side_effects)] // By design
-                fn mul_by_scalar<const CHECK: bool>(&self, rhs: &$t) -> Option<Self> {
-                    let rhs = Self::from_ref(rhs);
+                fn mul_by_scalar<const CHECK: bool>(&self, rhs: &$t) -> Option<Int<LIMBS2>> {
+                    if LIMBS2 < LIMBS {
+                        return None; // Cannot multiply if the left operand has fewer limbs than the output
+                    }
+                    let rhs: Int<LIMBS2> = Int::from_ref(rhs);
                     if CHECK {
-                        self.checked_mul(&rhs)
+                        rhs.checked_mul(&self.resize())
                     } else {
-                        Some(*self * rhs)
+                        Some(rhs * self.resize())
                     }
                 }
             }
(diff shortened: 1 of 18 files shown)


amit0365:
@codex review

frozenspider:
@codex review

zip-plus/src/pcs/utils.rs
```diff
@@ -16,16 +16,17 @@ fn err_too_many_variates(function: &str, upto: usize, got: usize) -> ZipError {
...
     {
-        // Inner ring should be at most+ 2*log_2(\rho^{-1}*d) + \log_2(d) +
-        // challenge_bits + eval_elem_bits - 1, where d is the size of the messages
-        // being encoded - so num_vars / 2
-        let d = div!(param_num_vars, 2);
+        // Inner ring should be at most 2*log_2(\rho^{-1}*d) + \log_2(d) +
+        // challenge_bits + eval_elem_bits - 1, where d is the log of the size
+        // of the messages being encoded (i.e. log2(row_len)).
+        let d = row_len.ilog2() as usize;
         let codeword_bits = ilog_round_up!(mul!(Lc::REPETITION_FACTOR, d), usize);
```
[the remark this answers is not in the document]
frozenspider: While that's a corner case that requires specifically `PnttConfigF65537_1_X<0>`, it's technically possible to specify parameters leading to this situation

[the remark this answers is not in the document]
frozenspider: ~~Having zero depth breaks other assumptions so disallowed zero depth via a compile-time check~~
Actually, paper permits depth-0 IPRS, so the fix should be different

utils/src/mul_by_scalar.rs
```diff
@@ -45,14 +45,17 @@ impl<const LIMBS: usize, const LIMBS2: usize> MulByScalar<&Int<LIMBS2>> for Int<
...
     ($($t:ty),*) => {
         $(
-            impl<const LIMBS: usize> MulByScalar<&$t> for Int<LIMBS> {
+            impl<const LIMBS: usize, const LIMBS2: usize> MulByScalar<&$t, Int<LIMBS2>> for Int<LIMBS> {
                 #[allow(clippy::arithmetic_side_effects)] // By design
-                fn mul_by_scalar<const CHECK: bool>(&self, rhs: &$t) -> Option<Self> {
-                    let rhs = Self::from_ref(rhs);
+                fn mul_by_scalar<const CHECK: bool>(&self, rhs: &$t) -> Option<Int<LIMBS2>> {
+                    if LIMBS2 < LIMBS {
+                        return None; // Cannot multiply if the left operand has fewer limbs than the output
```
amit0365: compile time `assert!` would be better here as both are `const`. Also the comment is backwards.

zip-plus/src/pcs/utils.rs
```diff
@@ -36,7 +37,7 @@ pub(super) fn validate_input<Zt: ZipTypes, Lc: LinearCode<Zt>, Pt>(
         let max_lc_bits = Zt::CombR::NUM_BITS;
         let actual_lc_bits: usize = add!(
             add!(
-                add!(mul!(codeword_bits, 2), ilog_round_up!(d, usize)),
+                add!(mul!(codeword_bits, 2), ilog_round_up!(d.max(1), usize)),
```
amit0365, line 42: A comment would be better here since `row_len = 1` is a valid configuration

amit0365 (changes requested):

zip-plus/src/pcs/phase_prove.rs
```diff
@@ -421,7 +428,7 @@ mod tests {
 
     #[test]
     fn prove_rejects_oversized_polynomial() {
-        let num_vars = 4;
+        let num_vars = 10;
         let (pp, _) = setup_test_params(num_vars);
         let oversized_poly: DenseMultilinearExtension<_> = (0..1 << 5).map(Int::from).collect();
```
amit0365: `num_vars` should be increased for the oversized poly

utils/src/mul_by_scalar.rs
frozenspider: Good point, this should be prevented at compile time

zip-plus/src/pcs/utils.rs
frozenspider: I don't follow - `row_len = 1` is valid, and it is properly supported now. Before this, it panicked at runtime.

amit0365: I meant since `row_len = 1` is a valid edge case, adding a comment about this would be helpful.

amit0365 (approved):

PR #158, opened 2026-03-28
Calculate and print proof sizes in benchmarks

frozenspider:
For this, we use our `Transcribable` trait that has been implemented for all proofs. More changes on this front:
* `Transcribable`/`ConstTranscribable` got their `GenTranscribable` common part extracted
* `Transcribable` bytes length prefix now defaults to u32 - 4 gigabytes should be plenty
* `delegate_transcribable!` macro defined to allow auto-implementation of transcribable traits for a single-field struct. Like `derive`, but does not require a separate proc-macro crate.
* Added several helpers like `(read/write)_transcription_bytes_subset` and `read_field_cfg`.
* PCS transcription `read/write` methods reworked to work with `Transcribable` instead of `ConstTranscribable`.
* All E2E tests now also check the consistency of proof (de)serialization.
* Additionally, discovered `BigLinear` UAIR witness being "too compressible". Changed the constraint it enforces from `col_i[next] = col_i[current]` to `col_i[next](1) = col_i[current](1)` to allow evaluations outside of X=1 be populated by random garbage (as long as we preserve number of 1-bits)

We also use `zstd` to compress our proof and print out the compressed size.
(I've also tried `gzip` and `brotli` but they gave worse proof sizes)

(Note that current design of `Transcribable`/transcript got complicated and we should probably pivot to make `serde` backend at some point, but for now this will do)


piop/src/sumcheck.rs
```diff
@@ -1,36 +1,115 @@
...
+    fn write_transcription_bytes_exact(&self, mut buf: &mut [u8]) {
+        buf = zinc_transcript::append_field_cfg::<F>(buf, &self.claimed_sum.modulus());
+        buf = {
+            let len =
+                u32::try_from(self.messages.len()).expect("messages length must fit into u32");
+            len.write_transcription_bytes_exact(&mut buf[0..u32::NUM_BYTES]);
+            &mut buf[u32::NUM_BYTES..]
+        };
+        for msg in &self.messages {
+            // This allows writing untrimmed polynomials without overhead
```
amit0365: Nice

zip-plus/src/pcs_transcript.rs
```diff
@@ -261,32 +273,43 @@ impl PcsVerifierTranscript {
...
+            T::read_num_bytes(&len_buf)
+        } else {
+            // LENGTH_NUM_BYTES == 0 means size is known at compile time via
+            // the ConstTranscribable blanket impl; read_num_bytes accepts an
+            // empty slice in that case.
+            T::read_num_bytes(&[])
+        };
+
+        read_stream_slice(&mut self.stream, data_len, |slice| {
+            Ok(T::read_transcription_bytes_exact(slice))
```
amit0365, line 298: `read_transcription_bytes_exact` can panic on malformed serialised proofs, making it as DOS vector for the verifier. Worth adding a TODO for `GenTranscribable` to return `Result` instead.

poly/src/univariate/dynamic/over_field.rs
```diff
@@ -389,6 +391,114 @@ impl<F: PrimeField> FromIterator<F> for DynamicPolynomialF<F> {
...
+where
+    F: PrimeField,
+    F::Inner: ConstTranscribable,
+    F::Modulus: ConstTranscribable,
+{
+    fn read_transcription_bytes_exact(bytes: &[u8]) -> Self {
+        if bytes.is_empty() {
+            return vec![].into();
+        }
+        let modulus_present = bytes[0] == 1;
```
amit0365: We should assert that bytes[0] flag is exactly 0/1. Otherwise invalid flag with empty polys will give an opaque error downstream.

amit0365 (approved):

PR #161, opened 2026-03-31
Custom row length in IPRS

### LINKED ISSUE #120: Customizable matrix sizes in Zip+ and support for IPRS codes for polynomials with i32 types

frozenspider:
Resolves #120, builds on top of #157

Reworked IPRS codes. Instead of having a bunch of constants in config, those values are now calculated at runtime to allow a configurable row length.
* A lot of values moved from constants in `Config` to fields in `Radix8PnttParams`.
* IPRS config is now specific to field only, and there's just one defined: `PnttConfigF65537`.
* Repetition factor (inverse rate) made a generic parameter on the IPRS code itself.

On top of that, the parameters themself were tweaked:
* Zip+ with IPRS now uses flat single-row matrix per polynomial in batch (i.e. before stacking said matrices for commitment).
* Added an alternative constructor for IPRS that uses a heuristics for depth: at least 1, and large enough for IPRS encoding matrix to have no more than 2^7 (128) columns.
  * This constructor is now used in all benchmarks as well as E2E tests.
* As a result, proving/verification time increased by up to 30% but proof sizes lowered drastically.

### DIFF
--- a/zip-plus/src/code/iprs.rs
+++ b/zip-plus/src/code/iprs.rs
@@ -36,6 +36,19 @@ where
         }
     }
 
+    /// Create a new IPRS code with the optimal depth heuristics trying to keep
+    /// number of columns in the base matrix small.
+    /// Currently, keeps number of columns <= 2^7 but this might be tweaked in
+    /// the future.
+    pub fn new_with_optimal_depth(row_len: usize) -> Self {
+        const MAX_BASE_COLS_LOG2: usize = 7;
+
+        let target_base_len = 1 << MAX_BASE_COLS_LOG2;
+        let depth = 1.max(((row_len / target_base_len).ilog2() as usize).div_ceil(3));
+
+        Self::new(row_len, depth)
+    }
+
     /// Encode without modular reduction, purely over the integers.
     fn encode_inner<In, Out>(&self, row: &[In]) -> Vec<Out>
     where
--- a/zip-plus/src/code/iprs.rs
+++ b/zip-plus/src/code/iprs.rs
@@ -4,18 +4,14 @@ use crate::{code::LinearCode, pcs::structs::ZipTypes};
 use crypto_primitives::{FromPrimitiveWithConfig, FromWithConfig};
 use num_traits::{CheckedAdd, CheckedMul};
 use pntt::radix8::params::Config as PnttConfig;
-pub use pntt::radix8::params::{
-    PnttConfigF65537_1_4, PnttConfigF65537_2_8, PnttConfigF65537_4_16, PnttConfigF65537_16_64,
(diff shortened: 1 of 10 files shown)


zip-plus/src/code/iprs.rs
```diff
@@ -4,35 +4,51 @@ use crate::{code::LinearCode, pcs::structs::ZipTypes};
...
     ops::{Add, AddAssign},
 };
-use zinc_utils::{from_ref::FromRef, mul, mul_by_scalar::MulByScalar};
+use zinc_utils::{from_ref::FromRef, mul_by_scalar::MulByScalar};
 
 /// Pseudo Reed-Solomon encoder over the integers. Internally uses a
 /// radix-8 NTT-style recursion with a base Vandermonde matrix sized
 /// `base_len x base_dim` (defaults to 64x32).
 ///
 /// To create a new instance, use [`IprsCode::default`] method.
```
amit0365: Stale comment from before

zip-plus/src/code/iprs.rs
```diff
@@ -4,35 +4,51 @@ use crate::{code::LinearCode, pcs::structs::ZipTypes};
...
+
+    /// Create a new IPRS code with the optimal depth heuristics trying to keep
+    /// number of columns in the base matrix small.
+    /// Currently, keeps number of columns <= 2^7 but this might be tweaked in
+    /// the future.
+    pub fn new_with_optimal_depth(row_len: usize) -> Self {
+        const MAX_BASE_COLS_LOG2: usize = 7;
+
+        let target_base_len = 1 << MAX_BASE_COLS_LOG2;
+        let depth = 1.max(((row_len / target_base_len).ilog2() as usize).div_ceil(3));
```
amit0365: Would this panic for small `row_len`?

amit0365 (approved):

zip-plus/src/code/iprs.rs
frozenspider: Good catch, thanks! It does panic indeed. Decided to make this a `Result<Code, ZipError>` instead.
Also added the (somewhat unrelated) test that IPRS code with minimal possible params row_len=1 and depth=0 works for Zip+.

Issue #107, opened 2026-02-13
General lookup argument


maksimryndin:
to avoid using a commitment scheme over the random finite field, batch the lookup claims and send the SOS decomposed chunks in the clear

maksimryndin:
@albert-garreta when you're done with your part, for the implementation could you pls re-assign to an engineer

frozenspider:
Use Plonky3's implementation: https://github.com/Plonky3/Plonky3/tree/main/lookup

PR #162, opened 2026-04-01
Add `multi_degree` sumcheck with lookup placeholder

amit0365:
Part of #107 — LogUp protocol needed to fully resolve.

## Summary

- **Lookup scaffold**:
  - Adds `LookupTableType` and `LookupColumnSpec` types to `zinc-uair`
  - Adds `lookup_specs` field in `UairSignature`
  - Defines all lookup proof/instance/intermediate types in `piop/src/lookup/structs.rs` 
  - All UAIRs pass `vec![]`; no prove/verify call sites yet — will be implemented with LogUp protocol

- **Multi-degree sumcheck**:
  - Adds `piop/src/sumcheck/multi_degree.rs` implementing `MultiDegreeSumcheck`
  - Runs G degree groups in lockstep with **one shared verifier challenge per round**, producing a common evaluation point
  - Needed to batch CPR (degree `max_deg+2`) and lookup groups into a single sumcheck
  - Adds `VerifierState::receive_round` to accept externally-provided challenges

> ⚠️ **Breaking change**: The multi-degree sumcheck absorbs additional metadata into the Fiat-Shamir transcript (`num_groups`, per-group degrees) compared to the previous single-group sumcheck. Proofs generated before this change will **not** verify.

## Test plan
- [x] `cargo test -p zinc-uair`
- [x] `cargo test -p zinc-piop` 
- [x] `cargo test -p zinc-protocol`
- [x] `cargo clippy -p zinc-uair -p zinc-piop`

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


uair/src/lookup_types.rs
```diff
@@ -0,0 +1,35 @@
...
+//! of [`UairSignature`] via `UairSignature::new(..., lookup_specs)`, the
+//! same way shifts are declared.
+
+/// Describes the type of lookup table a column should be checked against.
+/// Full table size = `2^width` for each type; decomposed into chunks of
+/// `chunk_width`
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum LookupTableType {
+    /// Binary polynomials of degree less than `width`, projected at element
+    /// `a`.
```
frozenspider: What is that element `a` being projected at mentioned here?

piop/src/combined_poly_resolver.rs
```diff
@@ -135,42 +142,66 @@ impl<F: InnerTransparentField + FromPrimitiveWithConfig + Send + Sync> CombinedP
...
+    /// - `transcript`: FS-transcript (absorbs `up_evals` and `down_evals`).
+    /// - `sumcheck_prover_state`: The CPR group's `ProverState` from
+    ///   `MultiDegreeSumcheck::prove_as_subprotocol` (states\[0\]).
+    /// - `ancillary`: Produced by [`prepare_sumcheck_group`]; carries column
+    ///   counts and `num_vars` needed to split the flat eval vector.
+    /// - `field_cfg`: Field configuration.
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn finalize_prover(
+        transcript: &mut impl Transcript,
+        sumcheck_prover_state: crate::sumcheck::prover::ProverState<F>,
```
frozenspider: `use crate::sumcheck::prover::{ProverState as SumcheckProverState}`

piop/src/sumcheck/verifier.rs
```diff
@@ -61,24 +61,30 @@ impl<F: FromPrimitiveWithConfig> VerifierState<F> {
...
         F::Inner: ConstTranscribable,
     {
+        let challenge: F = transcript.get_field_challenge(&self.config);
+        self.receive_round(prover_msg, challenge.clone());
+        challenge
+    }
+
+    /// Stores a challenge sampled by the verifier.
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn receive_round(&mut self, prover_msg: &ProverMsg<F>, challenge: F) {
```
frozenspider: Nit: Maybe `verify_round_with_challenge`? Also, the rustdoc comment above is more misleading than useful.

piop/src/lookup/structs.rs
```diff
@@ -0,0 +1,275 @@
...
+//!
+//! Proof types, prover/verifier intermediates, instance types, and error
+//! definitions live here. Specification types (`LookupTableType`,
+//! `LookupColumnSpec`) live in `zinc-uair` and are re-exported from the
+//! parent module.
+
+use std::marker::PhantomData;
+use thiserror::Error;
+use zinc_uair::LookupTableType;
+
```
frozenspider, line 11: Nit: Line breaks in `use` declarations confuse rustfmt - it never groups declarations between "blocks" and only appends them to the last block, so unless that's a desired behaviour (such as making `rayon` usage into its own group before other imports), we should group them together

piop/src/lookup/mod.rs
```diff
@@ -0,0 +1,7 @@
+//! Lookup argument for the Zinc+ PIOP.
+pub mod structs;
+
+pub use structs::*;
+
+// Re-export spec types from zinc-uair for convenience.
+pub use zinc_uair::{LookupColumnSpec, LookupTableType};
```
frozenspider: I would argue this is INconvenient - it's not obvious whether these types should be imported from `zinc_uair` or `zinc_piop`. Moreover, it's not obvious these are even the same types.

piop/src/lookup/structs.rs
```diff
@@ -0,0 +1,275 @@
...
+// ---------------------------------------------------------------------------
+
+/// Single-witness decomposition lookup instance (Phase 4a).
+///
+/// The witness is decomposed as:
+/// `w[i] = shifts[0]*chunks[0][i] + … + shifts[K-1]*chunks[K-1][i]`
+///
+/// All chunks look up into the same sub-table.
+#[derive(Clone, Debug)]
+pub struct DecompLookupInstance<F> {
```
frozenspider: Do we need a separate non-batched variant?

piop/src/lookup/structs.rs
```diff
@@ -0,0 +1,275 @@
...
+        /// Original trace column index.
+        column_index: usize,
+    },
+    /// Affine-combination lookup: parent eval = `Σ coeff·up_evals[col] +
+    /// offset`.
+    Affine {
+        /// `(column_index, coefficient)` pairs.
+        terms: Vec<(usize, i64)>,
+        /// Constant bit-polynomial offset encoded as a u32 bit pattern.
+        constant_offset_bits: u32,
```
frozenspider, line 59: Do we tailor all of this specifically for binary polynomials? How do we handle arbitrary polynomials and ints, or should we never encounter them here?

piop/src/lookup/structs.rs
```diff
@@ -0,0 +1,275 @@
...
+pub struct BatchedLookupProof<F> {
+    /// Per-group proofs, in group order.
+    pub group_proofs: Vec<BatchedDecompLogupProof<F>>,
+    /// Per-group metadata needed by the verifier.
+    pub group_meta: Vec<LookupGroupMeta>,
+}
+
+impl<F> BatchedLookupProof<F> {
+    /// Construct an empty proof (no lookup groups).
+    pub fn empty() -> Self {
```
frozenspider: Nit: Should be handled by `Default` instead

piop/src/lookup/structs.rs
```diff
@@ -0,0 +1,275 @@
...
+/// by their table type.
+///
+/// Columns with the same `LookupTableType` are batched into a single
+/// `BatchedDecompLogupProtocol` instance.
+pub fn group_lookup_specs(specs: &[zinc_uair::LookupColumnSpec]) -> Vec<LookupGroup> {
+    use std::collections::BTreeMap;
+
+    let mut map: BTreeMap<String, (LookupTableType, Vec<usize>)> = BTreeMap::new();
+    for spec in specs {
+        let key = format!("{:?}", spec.table_type);
```
frozenspider: We should not rely on debug representation of structs for anything other than debug printing!

piop/src/lookup/structs.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    #[error("multiplicity sum mismatch: expected {expected}, got {got}")]
+    MultiplicitySumMismatch { expected: u64, got: u64 },
+
+    #[error("final evaluation check failed")]
+    FinalEvaluationMismatch,
+
+    #[doc(hidden)]
+    #[error("internal")]
+    _Marker(PhantomData<F>),
```
frozenspider: Why do we need it? Maybe just remove `<F>` bound?

piop/src/sumcheck/multi_degree.rs
```diff
@@ -0,0 +1,592 @@
...
+}
+
+/// Proof for a multi-degree sumcheck.
+///
+/// `group_messages[g][round]` = prover message for group g in that round.
+/// All groups share verifier challenges, common evaluation point.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct MultiDegreeSumcheckProof<F> {
+    /// List of prover messages, one for each round per group.
+    group_messages: Vec<Vec<ProverMsg<F>>>,
```
frozenspider: Import `ProverMsg` as `SumcheckProverMsg` otherwise it's unclear what kind of prover message is that

frozenspider (changes requested):

piop/src/sumcheck/verifier.rs
amit0365: Good call. Edited the doc comments too. Let me know if they make sense.

uair/src/lookup_types.rs
amit0365: `a` doesnt belong here. Its referring the element we use to project the binary polys to get BitPoly. Removed

piop/src/lookup/mod.rs
amit0365: Removed.

piop/src/lookup/structs.rs
amit0365: My bad. This was sloppy.

piop/src/lookup/structs.rs
amit0365: No, but found out this is not needed as I am implementing LogUp so removed it with some other structs `BatchedDecompLookupInstance`, `LookupSumcheckGroup` and `LookupVerifierPreSumcheck` - these need to be modified. Better to add them in next PR.

piop/src/lookup/structs.rs
amit0365: This was sloppy again. Sorry! Using `LookupTableType` as key instead.

piop/src/lookup/structs.rs
amit0365: Yes, `Affine` combination is only used for Binary polynomials, specifically for SHA arithmetisation from @albert-garreta agentic-branch. Added a comment.

frozenspider (approved):
Great job!

PR #164, opened 2026-04-07
Better benchmarking for IPRS

frozenspider:
* Changed Zip (integer variant of Zip+) bechnamrks to use IPRS instead of RAA.
  * Widened codeword from `i64` to `i128` for that to work.
* In Zip+ common benches for encoding, use message sizes from 2^8 to 2^16 (was 2^7 to 2^10)
* Changed Zip+ benching code to skip a benchmark if linear code cannot be constructed for the given parameters.
  * This happens for IPRS since we can't construct it with more that 2^16 elements for the current F65537 config - meaning with rate 1/4, max message size is 2^14
* To avoid overflowing when encoding binary polynomials to `Z_{64}[X]`, changed IPRS depth heuristics from keeping encoding matrix columns <= 128 to <= 256. Note that this does slow down code chosen for some row lengths significantly due to lower depth being used, but this is okay for now
  * In the future, we'd like to have a check at the IPRS creation time, not at runtime
```
Zip+ IPRS/EncodeMessage: i32 -> i128, row_len=2048, rate=1/4, depth=2
                        time:   [319.00 µs 320.60 µs 322.43 µs]
Zip+ IPRS/EncodeMessage: i32 -> i128, row_len=2048, rate=1/4, depth=1
                        time:   [1.7413 ms 1.7529 ms 1.7675 ms]

Zip+ IPRS/EncodeMessage: i32 -> i128, row_len=16384, rate=1/4, depth=3
                        time:   [3.1971 ms 3.2233 ms 3.2516 ms]
Zip+ IPRS/EncodeMessage: i32 -> i128, row_len=16384, rate=1/4, depth=2
                        time:   [15.254 ms 15.441 ms 15.647 ms]
```
* Added `params_string()` to linear codes to report config parameters as string for benchmarking
* Added README section explaining how run benchmarks

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,32 @@
 
 WIP implementation and playground on Nethermind's improvement to Zinc SNARK .
 
+## Benchmarks
+
+Available benchmarks:
+
+| Benchmark          | What it measures                                                                                                                       |
+|--------------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| `zip_benches`      | PCS-level operations (encode, Merkle tree, commit, prove, verify) using **scalar** evaluations with IPRS codes. Uses `i32` evaluations |
+| `zip_plus_benches` | Same PCS-level operations using **polynomial** evaluations (degree 32 & 64) with both RAA and IPRS codes. Uses `{0,1}^D` evaluations.  |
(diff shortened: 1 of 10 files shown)


frozenspider:
@codex review

Issue #144, opened 2026-03-16
AI approach to optimizations


agarreta:
Ask AI to find optimization opportunities. If needed make it create flamegraphs to inform what optimizations seem more promising.

PR #171, opened 2026-04-11
Optimized mul_by_scalar

frozenspider:
Goes towards #144 

This improves E2E verifier running time by ~20-50% depending on the benchmark (with `unchecked` feature), both single-threaded and parallel.
This also improves prover times, although effect is less predictable.


frozenspider:
@codex review

PR #174, opened 2026-04-16
Step-by-step E2E rework

frozenspider:
Step-by-step approach introduced in #166 was reworked. Instead of having code duplication, E2E prover/verifier now structured as a state machine, where each step is only one possible for the current instance (except for MLE-first branching)
According to benches, running time was not affected by this change. 

Also, remaining `Air` structs were renamed to `Uair`


frozenspider:
@codex review

frozenspider:
@claude review

PR #176, opened 2026-04-21
E2E optimizations and tweaks

frozenspider:
1. In MLE-first approach, when the ideal is zero, the ideal check computation can be skipped - we know that the MLE evaluation is supposed to be 0, so P does not need to compute it: it just sends 0.
    * This only allows skippping MLE evaluation if ALL ideals are 0.
    * In combined polynomial ideal check, this was already handled but with more granularity - all zero ideal constraints were skipped while keeping non-zero ones.
2. Reworked MLE-first approach to handle linear constraints in MLE-first way while the rest are going through the regular combined subroutine. This allows using MLE-first approach for UAIRs with non-linear constraints.
3. Use IPRS code with rate 1/8 instead of 1/4, while also lowering number of column openings to 100 instead of 147.

Overall effect on E2E benchmarks (2^8 vars, parallel mode):
* Prover time (both combined and MLE-first) stayed pretty much the same
* Verifier time worsened by around to ~10%
* Proofs became smaller by ~30-40%, both compressed and uncompressed

### DIFF
--- a/piop/benches/ideal_check.rs
+++ b/piop/benches/ideal_check.rs
@@ -358,7 +358,7 @@ fn bench_big_linear_uair<const FIELD_LIMBS: usize>(
                         &mut transcript,
                         field_cfg,
                         project_trace_coeffs_column_major,
-                        prove_linear
+                        prove_mle_first
                     );
                     black_box(proof);
                 },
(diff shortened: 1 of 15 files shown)


frozenspider:
@codex review

PR #187, opened 2026-05-07
Chore split ub action

maksimryndin:
- [x] split ub action into several parallel jobs with the maximum timeout (to run every night by schedule). For the pcs part mostly happy paths are chosen - working here https://github.com/NethermindEth/zinc-plus/actions/runs/25451999523


frozenspider (approved):
Thanks!

PR #189, opened 2026-05-11
Folded Zip+ in PIOP

### LINKED ISSUE #139: Folded Zip+
Small variation of Zip+ that halves proof sizes (details to follow)

frozenspider:
Resolves #139 

* `UairTrace` now has two separate degrees, one for binary and one for arbitrary polynomials.
* For the initially provided trace, as well as public trace during the whole PIOP, those degrees are equal.
* P and V now do the folding trick, reducing the degree of binary polynomials by 4x (so 8 instead of 32).
* For P, steps are shifted, step 0 being the folding.
* Folding shortens proof but also worsens verifier time. To counteract it somewhat, IPRS heuristics has been lowered in order to increase depth.
  * Note that as the result, we can only encode polynomials with up to 12 variables.
  * Given that 4x folding increases number of variables by 2, we can no longer go above 10 variables in E2E benchmarks.

Technically, this trick could also be applied to integers, but it's not in this PR.

See #139 for the detailed description of the folding trick.

| # vars | `ShaProxy` bench  | Before  | After   | Change |
| ------ | ----------------- | ------- | ------- | ------ |
| 8      | P time (par)      | 23 ms   | 26 ms   | +15%   |
| 8      | V time (par)      | 3.2 ms  | 4.8 ms  | +50%   |
| 8      | Proof size (comp) | 218 KiB | 147 KiB | -30%   |
| 10     | P time (par)      | 64 ms   | 60 ms   | -5%    |
| 10     | V time (par)      | 9.5 ms  | 12 ms   | +25%   |
| 10     | Proof size (comp) | 258 KiB | 216 KiB | -15%   |

### DIFF
--- a/piop/benches/combined_poly_resolver.rs
+++ b/piop/benches/combined_poly_resolver.rs
@@ -56,7 +56,7 @@ fn bench_no_mult<const INT_LIMBS: usize, const FIELD_LIMBS: usize>(
     let max_degree = count_max_degree::<TestUairNoMultiplication<Int<INT_LIMBS>>>();
 
     let prove_cpr = |field_cfg: &<F<FIELD_LIMBS> as PrimeField>::Config,
-                     trace: &UairTrace<_, _, DEGREE_PLUS_ONE>,
+                     trace: &UairTrace<_, _, _, _>,
                      transcript: &mut Blake3Transcript| {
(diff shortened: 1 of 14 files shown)


frozenspider:
@codex review

Issue #185, opened 2026-05-06
Virtual columns (bit operations and linear combinations)


albert-garreta:

latifkasuli:
Hi, I’m interested in working on this since it blocks #91.

I read the relevant parts of the Zinc+ paper: Lemma 2.3, Remark 2.8, Remark 8.16, and the implementation notes in Section 9. My understanding is that this issue should add first-class UAIR support for virtual columns that are not committed directly, but whose MLE evaluations can be derived from committed/public columns.

Concretely, I think the useful initial scope is:

1. Extend `UairSignature` with virtual column specs for:
   - entry-wise bit-polynomial linear maps such as `ROTR`, `SHR`, possibly `SHL`;
   - affine/linear combinations of existing columns, including shifted columns.

2. Make the prover materialize these virtual evaluations where the constraint system needs them.

3. Make the verifier derive/check the same evaluations from the underlying committed/public column openings, so tampering with either the source or derived value is rejected.

4. Add tests around the SHA-relevant cases:
   - `SHR3(W)` and `SHR10(W)` from Remark 8.16;
   - linear-combination virtual columns for Ch/Maj-style booleanity checks.

Is that the intended scope for #185, or should bit-op virtual columns and affine-combination virtual columns be split into separate PRs?

albert-garreta:
Hi Latif!

Thank you for your interest! What you propose sounds good, we'll be happy if you help us out here. Feel free to split the scope into two parts :)

frozenspider:
@latifkasuli, please try to break your work onto small atomic PRs. Huge PRs are very hard to evaluate and review.

latifkasuli:
Thanks, agreed. I trimmed #191 down to a small foundation PR: UAIR metadata for bit-op virtual columns plus the reusable dense-polynomial `rot_c` / `shift_r_c` primitives.

The larger protocol implementation is preserved on a separate branch, but I will send it as follow-up PRs instead of keeping it all in #191. The next PR will handle the prover/verifier protocol plumbing and its e2e/tamper tests together.

PR #191, opened 2026-05-12
Add bit-op virtual column foundations

latifkasuli:
## Summary

- add UAIR metadata for bit-op virtual columns (`ROTR` / `SHR`) over binary polynomial cells
- validate bit-op source columns and counts against the declared binary polynomial cell width without storing that width in the signature
- add reusable dense-polynomial coefficient transforms for right-rotation and right-shift (`rotate_right` / `shr`)
- document the future down-row ordering invariant for shifted and bit-op virtual columns

## Context

Part of #185. This PR is the small foundation PR: it only adds the typed foundation needed by later protocol plumbing. It does not change prover/verifier behavior yet.

Follow-up PRs will wire these specs through CPR / multipoint evaluation and add the end-to-end SHA-relevant bit-op tests.

## Validation

- `.githooks/pre-push`
- `cargo test -p zinc-uair bit_op`
- `cargo test -p zinc-poly rotate_right`
- `cargo test -p zinc-poly shr`

### DIFF
--- a/poly/src/univariate/dense.rs
+++ b/poly/src/univariate/dense.rs
@@ -20,11 +20,13 @@ use std::{
 };
 use zinc_transcript::traits::{ConstTranscribable, GenTranscribable};
 use zinc_utils::{
+    add,
     from_ref::FromRef,
     inner_product::{InnerProduct, InnerProductError},
     mul_by_scalar::MulByScalar,
     named::Named,
     projectable_to_field::ProjectableToField,
+    rem,
 };
 
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


frozenspider:
@latifkasuli, a reminder: one this PR is ready for review, please move it out of draft phase

uair/src/lib.rs
```diff
@@ -182,7 +256,16 @@ pub struct UairSignature {
...
-    /// Column-type layout of the shifted (down) row.
+    /// Bit-op virtual column specs, in insertion order. Each spec references a
+    /// binary_poly source column and contributes one extra entry to the
+    /// binary_poly slice of the down row, appended after the shifted entries.
+    bit_op_specs: Vec<BitOpSpec>,
+    /// Cell width shared by binary_poly bit-op virtual columns.
+    ///
+    /// `None` iff no bit-op virtual columns are declared. When present, each
+    /// bit-op count is guaranteed to satisfy `0 < count < cell_width`.
+    binary_poly_cell_width: Option<usize>,
```
frozenspider: Since we pass cell width to `with_bit_op_specs` anyway, does it make sense to persist it in the signature?

poly/src/univariate/dense.rs
```diff
@@ -58,6 +58,68 @@ impl<R: Semiring + Zero, const DEGREE_PLUS_ONE: usize> DensePolynomial<R, DEGREE
...
+    /// endomorphism; unlike [`Self::rot_c`], `shift_r_c` cannot be viewed
+    /// as a map on `R[X]/(X^W - 1)`.
+    ///
+    /// Defined for any commutative ring `R` with a `Zero`; same dual use as
+    /// [`Self::rot_c`] across prover (bit-polynomial cells) and verifier
+    /// (lifted opening in `F_q^{<W}[X]`).
+    ///
+    /// Panics if `c == 0` or `c >= W`.
+    #[allow(clippy::arithmetic_side_effects)]
+    pub fn shift_r_c(&self, c: usize) -> Self {
```
frozenspider: I think we should align with established naming/convention from Rust core: [`rotate_right`](https://doc.rust-lang.org/core/intrinsics/fn.rotate_right.html) and [`impl Shr`](https://doc.rust-lang.org/beta/core/ops/trait.Shr.html).

Also, the comments are too wordy - there's no need to define what a `DensePolynomial` is on a method of that polynomial, this should be moved to the polynomial itself.

poly/src/univariate/dense.rs
```diff
@@ -58,6 +58,68 @@ impl<R: Semiring + Zero, const DEGREE_PLUS_ONE: usize> DensePolynomial<R, DEGREE
...
+    /// in the input (per Lemma 2.3) but is *not* a quotient-ring
+    /// endomorphism; unlike [`Self::rot_c`], `shift_r_c` cannot be viewed
+    /// as a map on `R[X]/(X^W - 1)`.
+    ///
+    /// Defined for any commutative ring `R` with a `Zero`; same dual use as
+    /// [`Self::rot_c`] across prover (bit-polynomial cells) and verifier
+    /// (lifted opening in `F_q^{<W}[X]`).
+    ///
+    /// Panics if `c == 0` or `c >= W`.
+    #[allow(clippy::arithmetic_side_effects)]
```
frozenspider: Here and above: The only side-effect op used is addition, and we have `add!` macro to avoid suppressing this warning

frozenspider (changes requested):

frozenspider (approved):
Thanks!

PR #192, opened 2026-05-14
Add bit-op virtual MLE projection helper

latifkasuli:
## Summary

- add `build_bit_op_virtual_mle` for projecting `BitOpSpec` virtual columns from row-major or column-major projected traces
- apply `Rot` / `ShR` to the source cell coefficients before evaluating at the projection point
- cover row-major, column-major, and invalid count cases with focused PIOP unit tests

## Context

Follow-up to #191 and part of #185. This is intentionally a small protocol-foundation PR: it adds the PIOP projection helper used by the later prover wiring, without changing prover/verifier behavior yet.

## Validation

- `.githooks/pre-push`
- `cargo test -p zinc-piop projections::tests -- --nocapture`

### DIFF
--- a/piop/src/projections.rs
+++ b/piop/src/projections.rs
@@ -8,10 +8,10 @@ use zinc_poly::{
     mle::DenseMultilinearExtension,
     univariate::dynamic::over_field::{DynamicPolyFInnerProduct, DynamicPolynomialF},
 };
-use zinc_uair::{Uair, UairTrace, collect_scalars::collect_scalars};
+use zinc_uair::{BitOp, BitOpSpec, Uair, UairTrace, collect_scalars::collect_scalars};
 use zinc_utils::{
-    UNCHECKED, cfg_extend, cfg_into_iter, cfg_iter, cfg_iter_mut, inner_product::InnerProduct,
-    powers,
+    UNCHECKED, add, cfg_extend, cfg_into_iter, cfg_iter, cfg_iter_mut, inner_product::InnerProduct,
+    powers, rem,
 };
 
 /// Row-indexed trace matrix: `trace[row][col]`.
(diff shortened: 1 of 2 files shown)


piop/src/projections.rs
```diff
@@ -312,6 +312,80 @@ pub fn evaluate_trace_to_column_mles<F: PrimeField + 'static>(
...
+    field_cfg: &F::Config,
+) -> DenseMultilinearExtension<F::Inner> {
+    let zero = F::zero_with_cfg(field_cfg);
+    let one = F::one_with_cfg(field_cfg);
+    let projection_powers: Vec<F> = powers(projecting_element.clone(), one, D);
+
+    let c = spec.op().count();
+    assert!(
+        c > 0 && c < D,
+        "BitOp count {} out of range for cell width D = {}",
```
frozenspider: Nit: Could save two lines by inlining `{c}` and `{D}`

piop/src/projections.rs
```diff
@@ -312,6 +312,80 @@ pub fn evaluate_trace_to_column_mles<F: PrimeField + 'static>(
...
+            BitOp::ShR(c) => (0..D)
+                .map(|i| {
+                    let j = add!(i, c);
+                    if j < D {
+                        coeffs[j].clone()
+                    } else {
+                        zero.clone()
+                    }
+                })
+                .collect(),
```
frozenspider: Since we've added `rotate_right` and `shr` to `DensePolynomial`, it probably makes sense to add those to `DynamicPolynomialF` as well, rather than describe those operations inline - it's weirdly inconsistent otherwise

frozenspider (changes requested):

frozenspider (approved):
Thank you!

Issue #91, opened 2026-02-10
Writing SHA-256 and ECDSA UAIRs (or porting from main-beta)


maksimryndin:
ask @albert-garreta for the details

latifkasuli:
Hi, I’m interested in taking this issue.

I’d prefer to implement this from `main` using the 2026/855 paper as the source of truth, not by directly relying on existing feature branches. My proposed split is:

1. Add the missing protocol support needed by the UAIRs: booleanity/range checks for `{0,1}<32[X]`, virtual bit-op handling for SHA rotations/shifts, and verifier-side public-structure checks.
2. Add a SHA-256 compression UAIR with e2e and tamper tests.
3. Add the ECDSA MSM UAIR for the secp256k1 double-and-add portion.
4. Add the combined SHA-256 + ECDSA UAIR.

Before I start, is this direction acceptable, and are there any constraints from `main-beta` or the paper implementation notes that you want followed?

PR #194, opened 2026-05-18
Splice bit-op virtuals in ideal check

latifkasuli:
## Summary
- store the declared binary-poly cell width on `UairSignature` so runtime materialization sites can apply bit ops with the right width
- add runtime-width `DynamicPolynomialF` bit-op helpers and use them from ideal-check
- splice bit-op virtuals into the ideal-check down row as `[shifted_binary, bit_op_binary, shifted_arbitrary, shifted_int]`
- add a mixed-splice test UAIR and exercise it through both ideal-check prover paths

Part of #185. This intentionally stops before CPR/mp-eval/protocol proof-wire plumbing.

## Notes
- The ideal-check test helpers now allow mixed-type signatures. The projection layer already supports binary_poly, arbitrary_poly, and int columns; the mixed-splice fixture needs that to catch binary/bit-op/arbitrary down-row ordering regressions.

## Validation
- `.githooks/pre-push`
- `cargo test -p zinc-piop ideal_check::tests::test_successful_verification`
- `cargo test -p zinc-test-uair test_constraint_degrees`
- `cargo test -p zinc-uair bit_op_specs`
- `cargo test -p zinc-piop projections::tests`

### DIFF
--- a/piop/src/ideal_check/combined_poly_builder.rs
+++ b/piop/src/ideal_check/combined_poly_builder.rs
@@ -13,9 +13,12 @@ use zinc_poly::{
     univariate::dynamic::over_field::DynamicPolynomialF,
     utils::{ArithErrors as PolyArithErrors, build_eq_x_r_vec},
 };
-use zinc_uair::{ColumnLayout, ConstraintBuilder, TraceRow, Uair, ideal::ImpossibleIdeal};
+use zinc_uair::{
+    BitOp, BitOpSpec, ColumnLayout, ConstraintBuilder, ShiftSpec, TraceRow, Uair, UairSignature,
+    ideal::ImpossibleIdeal,
+};
 use zinc_utils::{
-    cfg_into_iter, cfg_iter, from_ref::FromRef, inner_transparent_field::InnerTransparentField,
+    add, cfg_into_iter, cfg_iter, from_ref::FromRef, inner_transparent_field::InnerTransparentField,
 };
 
 /// Evaluate combined polynomial MLEs at `evaluation_point` for a selected
@@ -30,7 +33,7 @@ use zinc_utils::{
 ///
 /// `trace_matrix` is row-indexed: `trace_matrix[row][col]`.
 #[allow(clippy::arithmetic_side_effects)]
-pub fn evaluate_for_constraints<F, U>(
+pub fn evaluate_for_constraints<F, U, const DEGREE_PLUS_ONE: usize>(
     trace_matrix: &RowMajorTrace<F>,
     projected_scalars: &HashMap<U::Scalar, DynamicPolynomialF<F>>,
     num_constraints: usize,
@@ -57,21 +60,35 @@ where
     // `all_rows[row_idx][constraint_idx]` is a `DynamicPolynomialF<F>`:
     // the combined polynomial value of constraint `constraint_idx` at
     // trace row `row_idx`.
+    let bit_op_down_offset = bit_op_down_offset(&uair_sig);
     let mut all_rows: Vec<Vec<DynamicPolynomialF<F>>> = cfg_into_iter!(0..num_rows - 1)
         .map(|row_idx| {
             let up = &trace_matrix[row_idx];
 
-            let down: Vec<DynamicPolynomialF<F>> = uair_sig
-                .shifts()
-                .iter()
-                .map(|spec| {
-                    if row_idx + spec.shift_amount() < num_rows {
-                        trace_matrix[row_idx + spec.shift_amount()][spec.source_col()].clone()
-                    } else {
-                        DynamicPolynomialF::zero() // zero padding
-                    }
-                })
-                .collect();
+            // Build the down row in the canonical order:
+            //   [shifted_binary, bit_op_binary, shifted_arbitrary, shifted_int]
+            // (cf. `UairSignature::with_bit_op_specs`). Splicing bit-op
+            // virtuals into the binary_poly slice keeps `down` consistent
+            // with `down_layout`; appending at the tail would misalign
+            // constraints on mixed-type shift UAIRs.
+            let mut down: Vec<DynamicPolynomialF<F>> =
+                Vec::with_capacity(uair_sig.shifts().len() + uair_sig.bit_op_specs().len());
+
+            let shifts = uair_sig.shifts();
+            for spec in &shifts[..bit_op_down_offset] {
+                push_shifted_down_entry(&mut down, trace_matrix, row_idx, num_rows, spec);
+            }
+
+            for spec in uair_sig.bit_op_specs() {
+                let source = &trace_matrix[row_idx][spec.source_col()];
+                down.push(apply_bit_op_to_poly::<F, DEGREE_PLUS_ONE>(
+                    source, spec, field_cfg,
+                ));
+            }
+
+            for spec in &shifts[bit_op_down_offset..] {
+                push_shifted_down_entry(&mut down, trace_matrix, row_idx, num_rows, spec);
+            }
 
             evaluate_constraints_for_row::<F, U>(
                 up,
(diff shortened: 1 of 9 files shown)


albert-garreta:
Hi! Thanks Latif for this work. For the purposes of reviewing it, could you maybe share a short explanation of what are your overall plans towards getting the virtual column functionality fully implemented? Like what steps are you planning to take and how do you expect the final PR to look like?

latifkasuli:
> Hi! Thanks Latif for this work. For the purposes of reviewing it, could you maybe share a short explanation of what are your overall plans towards getting the virtual column functionality fully implemented? Like what steps are you planning to take and how do you expect the final PR to look like?

Thanks Albert. The overall goal here is to unblock #91: writing the SHA-256 and ECDSA UAIRs from the paper. #185 is the protocol/UAIR infrastructure needed before those UAIRs can be implemented cleanly, especially for SHA-style bit operations without committing every derived column as an independent witness.

I’m trying to keep the work as a small stacked sequence:

- #191: foundation for bit-op virtual columns (`BitOp`, `BitOpSpec`, signature metadata, polynomial rotate/shift helpers).
- #192: projected virtual-MLE helper for bit-op virtuals.
- #194: ideal-check support only, including the down-row ordering `[shifted_binary, bit_op_binary, shifted_arbitrary, shifted_int]` and a mixed-splice regression fixture.

Planned next steps:

1. Add the prover/verifier protocol plumbing for bit-op virtual columns:
   - prover materializes bit-op virtual MLEs from committed source columns,
   - CPR carries their claimed evaluations separately from ordinary shifted `down_evals`,
   - multipoint evaluation batches those claims,
   - verifier derives the final bit-op openings from the source lifted openings via Lemma 2.3, so virtual columns are not trusted as independent witnesses.

2. Add focused e2e/tamper coverage:
   - a small UAIR with `ShR` and `Rot` virtuals pinned to committed expected columns,
   - tamper tests for bit-op evals and source lifted openings.

3. Once bit-op virtuals are merged, use that machinery toward #91:
   - first SHA-256 UAIR pieces that need `ROTR` / `SHR`,
   - then the remaining SHA-256/ECDSA UAIR work.
   If affine/linear-combination virtual columns are still needed for the paper’s intended SHA/ECDSA shape, I’ll keep that as a separate follow-up rather than mixing it into the bit-op PRs.

The final shape I’m aiming for is: UAIRs declare virtual columns in the signature; the prover derives them from existing columns; and the verifier binds them back to the committed/public source openings. That should give #91 the virtual-column support it needs without inflating the SHA/ECDSA UAIR PRs with protocol plumbing.

piop/src/ideal_check/combined_poly_builder.rs
```diff
@@ -57,21 +59,55 @@ where
...
+            }
+
+            for spec in shifts_iter {
+                if row_idx + spec.shift_amount() < num_rows {
+                    down.push(
+                        trace_matrix[row_idx + spec.shift_amount()][spec.source_col()].clone(),
+                    );
+                } else {
+                    down.push(DynamicPolynomialF::zero());
+                }
```
frozenspider: This is an exact copy-paste of the code above. Please extract the common code, in a macro on inline helper function

piop/src/ideal_check/combined_poly_builder.rs
```diff
@@ -57,21 +59,55 @@ where
...
+            // Build the down row in the canonical order:
+            //   [shifted_binary, bit_op_binary, shifted_arbitrary, shifted_int]
+            // (cf. `UairSignature::with_bit_op_specs`). Splicing bit-op
+            // virtuals into the binary_poly slice keeps `down` consistent
+            // with `down_layout`; appending at the tail would misalign
+            // constraints on mixed-type shift UAIRs.
+            let mut down: Vec<DynamicPolynomialF<F>> =
+                Vec::with_capacity(uair_sig.shifts().len() + uair_sig.bit_op_specs().len());
+
+            let mut shifts_iter = uair_sig.shifts().iter();
```
frozenspider: Do you even need `iter` if you use it like this? Maybe just access elements by their indices?

frozenspider: Do you even need `iter` if you use it like this? Maybe just access elements by their indices?

uair/src/lib.rs
```diff
@@ -430,6 +441,12 @@ impl UairSignature {
         &self.bit_op_specs
     }
 
+    /// The degree bound `W` of the bit-polynomial cell module `R^{<W}[X]`.
+    /// Returns `None` when the signature has no bit-op specs.
+    pub fn binary_poly_cell_width(&self) -> Option<usize> {
+        self.binary_poly_cell_width
```
frozenspider: Is there a use case when we want customizable binary cell widths in UAIR? I.e. we want to trim/expand coefficients before doing rotation or shift

frozenspider (commented):
Sorry for the delay @latifkasuli, left comments

uair/src/lib.rs
latifkasuli: Clarified this in the field/accessor docs. The stored width is not intended as per-virtual-column customization; it preserves the UAIR binary-cell degree bound after projection/trimmed `DynamicPolynomialF` values lose that bound. That fixed width is what `Rot`/`ShR` need for the correct modulus/zero-padding when high coefficients are zero.

frozenspider: Do we need to do `coeffs.resize` then, or do we expect coeffs to be of the needed size already (in which case we can e.g. `assert`)?

frozenspider: Oh, and that `ZincTypes` are const-parameterized by `DEGREE_PLUS_ONE`, is this width different?

latifkasuli: Implemented this invariant explicitly now. `resize(width, zero)` is still needed because projected `DynamicPolynomialF` values may be trimmed, so `coeffs.len() < width` is valid and represents omitted high zero coefficients. But `coeffs.len() > width` should not happen, so `rotate_right_with_width` and `shr_with_width` now assert that before resizing, with regression tests for both cases.

For this PR, the stored `binary_poly_cell_width` is the runtime copy of the same binary-cell width that appears as `DEGREE_PLUS_ONE` before projection/type erasure. It is not intended to differ from the const width or customize individual bit-op virtuals.

frozenspider: Okay, got it, thanks for the clarification.
If that is the case, how about we revert to old const-width `rot`/`shr`, and instead make ideal check and combined poly builder to accept a `const DEGREE_PLUS_ONE: usize` parameter much like full PIOP prover and verifier do?
Feels weird to store and query to duplicate that data at runtime, given that we already know it at compile time. Besides, with the current code it's not obvious those are the same.

latifkasuli: Refactored this as suggested. `UairSignature` no longer stores or exposes `binary_poly_cell_width`, and `with_bit_op_specs` no longer takes a runtime width. The ideal-check API now carries `const DEGREE_PLUS_ONE: usize`, matching the full protocol path, and bit-op materialization uses `DynamicPolynomialF::rotate_right::<DEGREE_PLUS_ONE>` / `shr::<DEGREE_PLUS_ONE>`.

I kept the defensive `coeffs.len() <= D` asserts inside the const helpers, so trimmed-short values still get zero-padded before the operation, while oversized dynamic polynomials fail fast.

frozenspider (approved):
LGTM!

frozenspider:
@codex review

piop/src/ideal_check/combined_poly_builder.rs
```diff
@@ -57,21 +60,35 @@ where
...
+
+            let shifts = uair_sig.shifts();
+            for spec in &shifts[..bit_op_down_offset] {
+                push_shifted_down_entry(&mut down, trace_matrix, row_idx, num_rows, spec);
+            }
+
+            for spec in uair_sig.bit_op_specs() {
+                let source = &trace_matrix[row_idx][spec.source_col()];
+                down.push(apply_bit_op_to_poly::<F, DEGREE_PLUS_ONE>(
+                    source, spec, field_cfg,
```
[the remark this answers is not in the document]
frozenspider: That is fine as this PR's scope is ideal check only

Issue #186, opened 2026-05-06
Add support for having Z[X] and F_q[X] typed constraints simultaneously, for q a large prime, with two "branching proofs"


albert-garreta:
The following is a plan that Claude wrote. It explains (for the most part) what this task means.

# Dual-prime extension to Zinc+ protocol (`Folded 4×` path)

## Context

Today the `Folded 4×` protocol commits over `Z[X]`, then projects to `F_p[X]`
where `p` is the secp256k1 base prime (compile-time fixed in
[protocol/src/fixed_prime.rs](protocol/src/fixed_prime.rs)). All constraints
flow through that single projection.

The user wants per-constraint typing: each constraint is tagged either
`Z` or `F_p`. After commit, the verifier sends `p` (still secp256k1, fixed)
**and** a fresh prime `q` (drawn from Fiat–Shamir). The protocol then runs
two parallel branches:

- **`F_p`-branch**: every `F_p`-typed constraint is projected and checked
  via the small random `q` (this is the "usual" path of the original
  pre-fixed-prime protocol).
- **`Z`-branch**: every `Z`-typed constraint is projected and checked via
  the fixed `p` (current behavior, repurposed for the `Z`-branch only).

Both branches share randomness (`r*`, `r_0`) sampled in `[0, min{p,q}-1]`.
After step 5 each branch produces evaluation claims at the same `r_0`
for the columns it touched; some columns may appear in both. A single
joint batched Zip+ open at `r_0` discharges all column claims.

For the initial deliverable the user wants every `ShaEcdsaUair` constraint
tagged `F_p`, so the Z-branch is **wired but empty** in the bench. The
existing `prove_folded_4x` is preserved; a new
`prove_folded_4x_dual_prime` lives alongside it with a new bench group.

Decision on the tagging API: **`assert_in_ideal_typed` on the existing
single Ideal type**. Two-Ideal-types (`IdealZ` / `IdealFp`) is rejected as
too invasive — `Sha256Ideal<R>` already supports both rings, so attaching
a `ConstraintRing` enum tag per constraint slot is enough and avoids
splitting the constrain method.

---

### 1. Constraint typing — UAIR layer

In [uair/src/lib.rs](uair/src/lib.rs):

- Add `pub enum ConstraintRing { Z, Fp }` (Copy, Eq).
- Extend `ConstraintBuilder` with `assert_in_ideal_typed(&mut self, expr,
[...]

PR #198, opened 2026-05-26
uair: add typed constraint tags to ideal collection

peter941221:
This PR adds the first typed-constraint plumbing slice for #186.

It keeps the change at the UAIR boundary. The patch introduces `ConstraintRing`, adds a defaulted `assert_in_ideal_typed(...)` entrypoint on `ConstraintBuilder`, and records per-constraint tags in `IdealCollector`.

I split this out because the full dual-prime work spans `uair`, `piop`, and `protocol`, while this metadata plumbing is the smallest reusable piece and the easiest part to review independently. Existing builders keep their current behavior because the new typed entrypoint falls back to `assert_in_ideal(...)` by default.

This PR does not change `piop` routing or prover/verifier branching yet.

Validation:
`cargo test -p zinc-uair`
`cargo test -p zinc-uair --all-features`
`cargo check -p zinc-piop`
`cargo check -p zinc-protocol`

### DIFF
--- a/uair/src/ideal_collector.rs
+++ b/uair/src/ideal_collector.rs
@@ -1,7 +1,7 @@
 use zinc_utils::from_ref::FromRef;
 
 use crate::{
-    ConstraintBuilder, TraceRow, Uair,
+    ConstraintBuilder, ConstraintRing, TraceRow, Uair,
     dummy_semiring::DummySemiring,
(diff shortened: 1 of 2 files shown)


peter941221:
Thanks for the clarification. I’m closing this PR so you can take the feature in the direction you want.

Issue #179, opened 2026-05-01
Improve projected scalars cache lookup


frozenspider:
(Based on https://github.com/NethermindEth/zinc-plus/commit/6517386a764d71f6b5e2cc3dbd2971a28a1c8321)

Projected scalar cache is currently keyed by `U::Scalar` - so `DensePolynomial<R, 32>`. Hashing it is somewhat expensive and hashmap lookups need to be done repeatedly. Investigate opportunities to avoid hashing it.

frozenspider:
I don't see any performance with 6517386 when rebased on top of current `main` (ff64869), I also tried different caching approach, so closing this for now.
If needed, cache introduction would be simplified by #199

PR #199, opened 2026-05-27
Abstract away scalar projection HashMap

frozenspider:
This is a small cleanup introducing typesafe wrapper around scalar projection that incapsulates underlying hashmap.
If needed, we can later add caching or change key resolution there (see #179), but for now this doesn't seem to be beneficial.


frozenspider:
@codex review

PR #200, opened 2026-05-31
Optimize prove_f row products

Kuhai9801:
## Summary
- lift `q_1` into `Zt::CombR` once when the conversion is safe
- compute per-row products in the integer ring instead of repeated field conversion
- fall back to the existing field inner product path when conversion is unavailable or checked accumulation overflows
- add focused coverage for direct `prove_f` usage and field-element integer parsing

## Verification
- `cargo fmt -p zip-plus --all`
- `cargo test -p zip-plus --lib phase_prove -- --nocapture`
- `cargo test -p zip-plus --lib phase_verify -- --nocapture`
- `cargo check -p zinc-protocol`

### DIFF
--- a/zip-plus/src/pcs/phase_prove.rs
+++ b/zip-plus/src/pcs/phase_prove.rs
@@ -7,9 +7,9 @@ use crate::{
     },
     pcs_transcript::PcsProverTranscript,
 };
-use crypto_primitives::{FromWithConfig, IntoWithConfig, PrimeField};
+use crypto_primitives::{ConstIntRing, FromWithConfig, IntoWithConfig, PrimeField};
 use itertools::Itertools;
-use num_traits::{ConstOne, ConstZero, Zero};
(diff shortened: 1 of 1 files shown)


albert-garreta:
@Kuhai9801 thanks for the PR! Do you know how much it improves the prover time?

1sgtpepper:
Thanks @albert-garreta for asking for concrete prover-time numbers here. I opened #201 as a narrower replacement that uses the one-hot structure of Boolean equality tensors in `ZipPlus::prove_f` and keeps the dense fallback for non-Boolean points.

The PR body includes Criterion results from the existing Zip+ prove benchmarks. On `BPoly<63>`, the measured speedups are 1.28x for RAA batch 1 at `2^16`, 1.36x for RAA batch 5 at `2^16`, and 4.59x for IPRS batch 1 at `2^13`.

PR #202, opened 2026-06-02
perf(piop): avoid clones in virtual bit-op evals

wu-s-john:
Use borrowed/in-place field addition while accumulating rotated and shifted binary polynomial evaluations. These paths run once per relevant set bit across each virtual bit-op or shifted bit-slice column, so avoiding clones removes a large amount of temporary field-element copying without changing the immediate-reduction semantics.

### DIFF
--- a/test-uair/src/sha_ecdsa.rs
+++ b/test-uair/src/sha_ecdsa.rs
@@ -80,15 +80,17 @@ use zinc_uair::{
 
 use crate::{
     GenerateRandomTrace,
-    ecdsa::{self, FINAL_ROW as ECDSA_FINAL_ROW, NUM_SHAMIR_ROUNDS},
+    ecdsa,
     ecdsa_doubling::{EC_FP_INT_LIMBS, EcdsaFpRing},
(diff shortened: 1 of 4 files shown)


frozenspider (approved):
Thanks!

test-uair/src/sha_ecdsa.rs
```diff
@@ -80,15 +80,17 @@ use zinc_uair::{
...
     sha256::{self, Sha256CompressionSliceUair, Sha256Ideal},
 };
+#[cfg(test)]
+use crate::ecdsa::FINAL_ROW as ECDSA_FINAL_ROW;
 
 use crypto_primitives::crypto_bigint_int::Int;
 
 // Re-export for convenience.
-pub use crate::ecdsa::FINAL_ROW;
+pub use crate::ecdsa::{FINAL_ROW, NUM_SHAMIR_ROUNDS};
```
frozenspider: I don't think we need it in this PR

wu-s-john: Addressed. Please merge the code

frozenspider:
Whoops, this was merged to a wrong branch 😄

PR #203, opened 2026-06-03
Non-recursive `eq` polynomial builder

frozenspider:
Use iterative formula for `build_eq_x_r` instead of a recursive approach.
This leads to ~10-40% speedup depending on the UAIR and number of variables.
Also, moved relevant test from `dense` module closer to `utils`.


frozenspider:
Thanks @wu-s-john for the pointer!

PR #204, opened 2026-06-03
Plumb bit-op virtuals through CPR

latifkasuli:
## What changed

This is the next small follow-up after #191, #192, and #194 for #185, which blocks #91.

It extends the combined polynomial resolver (CPR) path to carry bit-op virtual column evaluations separately from row-shift down evaluations:

- Adds `bit_op_evals` to the CPR proof and verifier subclaim.
- Adds `bit_op_down_mles` input to `CombinedPolyResolver::prepare_sumcheck_group`.
- Splices bit-op MLEs into the CPR sumcheck down-row layout as:
  `[shifted_binary..., bit_op_binary..., shifted_arbitrary..., shifted_int...]`.
- Keeps the wire format explicit: row-shift evals remain in `down_evals`, bit-op virtual evals travel in `bit_op_evals`.
- Updates current no-bit-op protocol and benchmark callers with `Vec::new()` placeholders.

This PR does not bind `bit_op_evals` to source openings yet. That remains for the next atomic PR in the multipoint-eval/protocol verifier layer. The CPR proof values introduced here are therefore plumbing only until that follow-up lands.

## Validation

- `cargo test -p zinc-piop combined_poly_resolver`
- `cargo check -p zinc-protocol`
- `.githooks/pre-push`

### DIFF
--- a/piop/src/combined_poly_resolver/structs.rs
+++ b/piop/src/combined_poly_resolver/structs.rs
@@ -8,13 +8,23 @@ use crate::combined_poly_resolver::CombinedPolyResolverError;
 ///
 /// Note: the sumcheck proof now lives at the protocol
 /// level as part of `MultiDegreeSumcheckProof`.
+///
+/// `bit_op_evals` is the prover's claim about the bit-op virtual columns'
+/// MLE evaluations at the shared CPR point. These are *not* trusted by
+/// themselves: they are bound back to the source columns' lifted openings at
+/// the multi-point evaluation endpoint via Lemma 2.3.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Proof<F: PrimeField> {
     /// The evaluation of the projected trace columns MLEs at the shared point.
     pub up_evals: Vec<F>,
     /// The evaluations of the shifted projected trace columns MLEs at the
-    /// shared point.
+    /// shared point. Carries *only* the row-shift virtual columns —
+    /// bit-op virtuals are sent separately in `bit_op_evals` to avoid
+    /// overloading shift semantics on the wire.
     pub down_evals: Vec<F>,
+    /// The evaluations of the bit-op virtual columns (ROTR / SHR) at the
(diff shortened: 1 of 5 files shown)


piop/src/combined_poly_resolver/structs.rs
```diff
@@ -46,21 +59,28 @@ where
...
         )
     }
 }
 
 impl<F: PrimeField> Proof<F> {
-    /// Check if `up_evals` and `down_evals` vectors have the expected lengths.
+    /// Check that the proof's evaluation vectors have the expected lengths.
+    ///
+    /// `num_down_cols` counts the *shift* virtual columns only; bit-op
+    /// virtuals are tracked separately via `num_bit_op_specs`.
```
frozenspider: I think this bit is unnecessary and follows from the function signature

piop/src/combined_poly_resolver/structs.rs
```diff
@@ -87,12 +114,20 @@ pub struct ProverState<F: PrimeField> {
...
+    /// Number of shift-virtual (down) columns.
     pub num_down_cols: usize,
+    /// Number of bit-op virtual columns. These sit inside the binary_poly
+    /// slice of the down row, after the shifted-binary entries and before
+    /// any non-binary down entries (see `UairSignature::with_bit_op_specs`).
+    pub num_bit_op_specs: usize,
+    /// Offset (within the down portion of the post-sumcheck eval vec) at
+    /// which the bit-op evals start. Equivalent to "number of shifts whose
+    /// source column is in the binary_poly range."
+    pub bit_op_down_offset: usize,
```
frozenspider: This directly follows from the UAIR signature, I don't think this should be part of the proof. You could make this into a helper method in `UairSignature` instead, similarly to `dummy_rows`, etc.

frozenspider (changes requested):

frozenspider:
@codex review

frozenspider (approved):

PR #205, opened 2026-06-04
Delayed Modular Reduction for Fast Polynomial Evaluations

wu-s-john:
This PR speeds up binary polynomial evaluation under the scalarization challenge by using delayed modular reduction for the hot equality-weight accumulations.

Delayed modular reduction means we do not reduce modulo the field prime after every field addition. Instead, we accumulate several field elements in a wider integer representation that is large enough to hold the unreduced sum, then reduce the accumulator back into the field once at the end.

For a binary column with row bits c_{b,i}, evaluation has the shape:

Σ_b eq_r(b) · (Σ_i c_{b,i} α^i)

Equivalently:

Σ_i α^i · (Σ_b eq_r(b) c_{b,i})

This PR accelerates the inner coefficient sums:

Σ_b eq_r(b) c_{b,i}

Those sums are just conditional additions of cached eq_r(b) values for set bits. Instead of performing a modular reduction after each selected addition, we accumulate the Montgomery limbs in a 5-limb Uint<5> delayed-reduction accumulator and perform one Barrett reduction at the end.

### DIFF
--- a/piop/src/combined_poly_resolver.rs
+++ b/piop/src/combined_poly_resolver.rs
@@ -125,7 +125,7 @@ where
                             continue;
                         }
                         if let Some(w) = &weights[i] {
-                            acc += w.clone();
+                            acc += w;
                         }
                     }
                     acc.into_inner()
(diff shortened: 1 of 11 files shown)


frozenspider:
Hey @wu-s-john, could you please update the branch by resolving conflicts?

PR #212, opened 2026-06-11
Adopt next iteration of crypto primitives

frozenspider:
Adopt changed introduced in https://github.com/NethermindEth/crypto-primitives/pull/33.
This enables lifting field elements to integers - attempt to do that was done in #200.
Instead of writing `field.inner` to transcription, write `field.lift_to_integer`. This has no observable impact on UAIR benchmark performance.


frozenspider:
@codex review

PR #211, opened 2026-06-10
Batch bit-op virtual claims in multipoint eval

latifkasuli:
## What changed

This is the next atomic follow-up after #191, #192, #194, and #204 for #185, which blocks #91.

This PR extends the multipoint-eval subprotocol so bit-op virtual claims can be batched into the final consistency check:

- Adds separate bit-op batching challenges to the mp-eval subclaim.
- Lets the prover include bit-op virtual MLEs in the precombined MLE.
- Lets the verifier include CPR-provided bit-op evals in the expected sum.
- Extends the final subclaim check to fold verifier-derived bit-op open evals into the final check.
- Keeps current protocol and bench call sites as no-op empty bit-op slices until the next PR materializes and derives real bit-op openings.

## What is intentionally deferred

This PR does not yet materialize bit-op virtual MLEs in the full protocol prover, and it does not derive bit-op openings from lifted source openings in the protocol verifier. That is the next atomic PR. Existing UAIRs continue to use empty bit-op slots.

## Validation

- Red/green: cargo test -p zinc-piop bit_op_virtual_opening_is_bound_in_subclaim
- cargo test -p zinc-piop multipoint_eval
- cargo check -p zinc-protocol
- .githooks/pre-push

PR #220, opened 2026-06-23
Wire bit-op virtuals through protocol constraint families

latifkasuli:
This is the next atomic follow-up after #191, #192, #194, #204, and #211 for #185, which blocks #91. It closes the executable protocol path for bit-op virtual columns across both Q[X] and F_q[X] constraint families.

## What changed

- Prover: materialize bit-op virtual MLEs from committed source columns for the Q[X] family and each declared F_q[X] family, using the matching projecting element and field config per family.
- Verifier: derive bit-op open evaluations from the public UAIR signature and the corresponding source lifted openings via the rotate/shift polynomial operations, then feed those derived values into multipoint-eval verification.
- Tests: add `TestUairBitOpsFqFamily`, which constrains `ShR(w, 3)` through both Q[X] and F_q[X], plus e2e and tamper regression coverage.

## Coverage note

The e2e tamper tests in this PR are intentionally framed as end-to-end regressions:

- `test_bit_ops_fq_family_tamper_source_lifted_evals` perturbs the source lifted eval, so it changes both the committed-source opening and the verifier-derived bit-op opening.
- `test_bit_ops_fq_family_tamper_bit_op_eval` perturbs the F_q-family bit-op CPR claim; that is caught by CPR constraint reconstruction before the final multipoint-eval check.

The isolated bit-op multipoint-eval binding coverage lives in #211 (`bit_op_virtual_opening_is_bound_in_subclaim`), and the CPR-level integrity for bit-op evals lives in #204. The honest e2e here exercises the new verifier derivation path because the prover and verifier compute the bit-op virtual through different routes that must agree.

This completes the bit-op virtual-column protocol chain. Affine / linear-combination virtual columns for Ch/Maj-style shapes remain a separate follow-up part of #185, and SHA-256/ECDSA UAIR adoption follows in #91.

## Validation

Bit-op / F_q regression tests covered by `cargo test -p zinc-protocol`:

- `test_e2e_bit_ops_with_fq_family`
- `test_bit_ops_fq_family_tamper_source_lifted_evals`
- `test_bit_ops_fq_family_tamper_bit_op_eval`

Commands run locally:

```text
cargo test -p zinc-protocol
.githooks/pre-push
```

### DIFF
--- a/protocol/src/verifier.rs
+++ b/protocol/src/verifier.rs
@@ -19,7 +19,7 @@ use zinc_transcript::{
     traits::{ConstTranscribable, Transcript},
 };
 use zinc_uair::{
-    Uair, UairSignature, UairTrace,
+    BitOp, Uair, UairSignature, UairTrace,
     constraint_counter::count_constraints,
     ideal::{Ideal, IdealCheck},
     ideal_collector::IdealOrZero,
@@ -258,6 +258,7 @@ pub struct VerifierSumchecked<
     /// CPR subclaim's evaluation point ($r^\star$)
     cpr_eval_point: Vec<F>,
     cpr_up_evals: Vec<F>,
+    cpr_bit_op_evals: Vec<F>,
     cpr_down_evals: Vec<F>,
     /// Per-prime CPR subclaim evaluation points (r*, lifted into
     /// each family's field). Length `n_fq`. Empty for UAIRs with no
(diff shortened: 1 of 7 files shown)


protocol/src/verifier.rs
```diff
@@ -1556,6 +1576,27 @@ impl<IdealOverF: Ideal> VerifierPcsVerified<IdealOverF> {
...
+    field_cfg: &F::Config,
+) -> Result<Vec<F>, ProtocolError<F>> {
+    specs
+        .iter()
+        .map(|spec| {
+            let source = &all_lifted[spec.source_col()];
+            let transformed = match spec.op() {
+                BitOp::Rot(c) => source.rotate_right::<D>(c, field_cfg),
+                BitOp::ShR(c) => source.shr::<D>(c, field_cfg),
+            };
```
frozenspider: This seems to be a recurring pattern, this is 3rd time this exact code block appears. How about we define something like `BitOp::transform<F>(&self, &DynamicPolynomialF<F>, &F::Config) -> DynamicPolynomialF<F>` and reuse it across the codebase?
(I don't think it makes sense to make it more generic than `DynamicPolynomialF<F>` at this point)

frozenspider (approved):
Thanks!

PR #221, opened 2026-06-24
Add affine virtual column specs

latifkasuli:
Follow-up for #185, which blocks #91. This starts the linear/affine-combination half after the bit-op virtual-column chain (#191, #192, #194, #204, #211, #220).

This PR is intentionally only the `zinc-uair` foundation: it lets a UAIR signature describe affine virtual booleanity targets, but does not materialize them or change any proof path yet.

## What changed

- Added `AffineVirtualTerm`, representing one signed term over a binary-polynomial source column, with optional forward row shift using the same convention as `ShiftSpec`.
- Added `AffineVirtualSpec`, representing an expression of the form `ones_coefficient * 1_D + sum_i coefficient_i * source_i[row + shift_i]`.
- Added `UairSignature::with_affine_virtual_specs(...)` and `affine_virtual_specs()`.
- Validates that every affine virtual source references a `binary_poly` column.

## Design notes

Affine virtuals are Q-side booleanity targets for expressions such as the Ch/Maj lookup constraints from the paper, e.g. `a + b + c - 2m` and `(1_D - e) + g - 2u`.

They are not committed columns and they do not add entries to `down_cols`. The existing `down`/CPR bit-op virtual path is deliberately untouched.

- Building affine virtual bit-slice MLEs for the booleanity sumcheck.
- Extending booleanity accounting so virtual booleanity targets are counted separately from committed witness binary columns.
- Binding affine virtual alpha-prime bridge claims back to committed source openings.
- SHA-256 Ch/Maj adoption.

Those will be separate small PRs so this remains reviewable.

## Validation

- `cargo test -p zinc-uair`
- `cargo +nightly fmt --all -- --check`
- `cargo clippy -p zinc-uair --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/uair/src/lib.rs
+++ b/uair/src/lib.rs
@@ -195,6 +195,115 @@ impl BitOpSpec {
 }
 
 // ---------------------------------------------------------------------------
+// Affine virtual columns for booleanity targets
+// ---------------------------------------------------------------------------
+
+/// One term in an affine virtual binary-polynomial expression.
+///
+/// `source_col` uses the same flat `binary_poly || arbitrary_poly || int`
+/// indexing as [`ShiftSpec`]. [`UairSignature::with_affine_virtual_specs`]
+/// validates that every source is a `binary_poly` column.
+///
+/// `row_shift = 0` means the current row. Positive shifts use the same
+/// zero-padded row access convention as [`ShiftSpec`]: row `i` reads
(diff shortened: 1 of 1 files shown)


frozenspider:
@codex review

frozenspider (approved):
LGTM!

PR #222, opened 2026-06-26
Add affine virtual booleanity MLEs

latifkasuli:
Next atomic follow-up after #221 for #185, which blocks #91.

## What changed

- Adds `build_affine_virtual_bit_slice_mles` for materializing affine virtual residual bit-slice MLEs from the full binary-polynomial trace in public-then-witness UAIR order, including shifted terms and `1_D` offsets.
- Adds `prepare_affine_virtual_sumcheck_group`, which routes affine virtual targets through a generic degree-3 booleanity group without the `BinaryPoly` round-1 fast path.
- Reuses the committed booleanity path transcript challenge and zerocheck setup while keeping affine residual MLE materialization separate.
- Adds focused tests for residual materialization, shifted/ones offsets, public-prefix indexing, precise missing-input errors, honest and non-boolean roundtrips, and tampered proof rejection.

## Intentionally deferred

- Protocol integration and signature-derived booleanity target accounting are deferred until their first consumer; affine virtuals are not yet included in protocol proofs.
- The alpha-prime bridge that binds affine virtual bit-slice claims back to committed source openings is deferred to the soundness-closing follow-up.
- SHA-256 Ch/Maj adoption remains a later PR after the protocol wiring is merged.

## Validation

- `./.githooks/pre-push`

### DIFF
--- a/piop/src/lookup/booleanity.rs
+++ b/piop/src/lookup/booleanity.rs
@@ -1,11 +1,13 @@
 //! Booleanity (binary-polynomial lookup) argument.
 //!
 //! Proves that every coefficient of every witness binary-polynomial column is
-//! a bit $\in \\{0,1\\}$. The argument is structured as a single
-//! [`MultiDegreeSumcheckGroup`] of degree 3, batched alongside the existing
-//! CPR group with shared randomness, and emits bit-slice claims at the
+//! a bit $\in \\{0,1\\}$. The committed-witness path is structured as a
+//! [`MultiDegreeSumcheckGroup`] of degree 3, batched alongside the existing CPR
+//! group with shared randomness, and emits bit-slice claims at the
 //! multi-degree sumcheck output point $r^\star$ for the protocol layer's
 //! $\alpha'$ bridge into multipoint-eval (see "Soundness bridge" below).
+//! Affine virtual booleanity targets use a generic degree-3 group without the
+//! `BinaryPoly` round-1 fast path.
 //!
 //! # Relation
 //!
@@ -89,6 +91,7 @@ use zinc_transcript::{
     delegate_transcribable,
     traits::{ConstTranscribable, Transcript},
 };
+use zinc_uair::AffineVirtualSpec;
 use zinc_utils::{add, mul, powers};
 
 //
@@ -98,14 +101,16 @@ use zinc_utils::{add, mul, powers};
 /// Booleanity sumcheck group constructor / verifier.
 pub struct BooleanityChecker<C: SemiringConfig>(PhantomData<C>);
 
+type BooleanityZerocheckSetup<F> = (Vec<F>, Vec<F>, CombFn<F>);
+
 /// Proof produced by the booleanity prover.
 ///
-/// Carries `bit_slice_evals`: per-column bit-slice MLE evaluations at the
-/// multi-degree sumcheck output point $r^\star$, flat in
-/// `(j-major, i-minor)` order with length `num_wit_bin_cols * D`. It is
-/// absorbed into the transcript by `finalize_prover` / `finalize_verifier`,
-/// after which the protocol layer squeezes the $\alpha'$ challenge
-/// used to bind the bit-decomposition into the multipoint-eval boundary.
+/// Carries `bit_slice_evals`: per-booleanity-target bit-slice MLE evaluations
+/// at the multi-degree sumcheck output point $r^\star$, flat in
+/// `(j-major, i-minor)` order. It is absorbed into the transcript by
+/// `finalize_prover` / `finalize_verifier`, after which the protocol layer
+/// squeezes the $\alpha'$ challenge used to bind the bit-decomposition into
+/// the multipoint-eval boundary.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct BooleanityProof<F> {
     /// Flat list of `\widetilde{v_{j,i}}(r*)`, ordered `(j-major, i-minor)`.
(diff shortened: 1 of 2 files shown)


frozenspider:
Hey @latifkasuli! Sorry, our priorities were shifted for a while, but now we're once again targeting merging virtual columns in a near future. Would you be up to updating this PR for review and merge? Thanks!

frozenspider:
@codex review

piop/src/lookup/booleanity.rs
```diff
@@ -685,6 +700,174 @@ where
...
+            let eq_r_val = &mle_values[0];
+            let sum = batched_booleanity_sum(&field_cfg, &mle_values[1..], &alpha_powers);
+            field_cfg.mul(&sum, eq_r_val)
+        })
+    };
+
+    (r, alpha_powers, comb_fn)
+}
+
+fn prepare_generic_bit_slice_sumcheck_group<C>(
```
frozenspider: This is called `prepare_generic...` but it's only used in `prepare_affine_virtual_sumcheck_group`. Are there plans to use it somewhere else? If not, I think it should be inlined, as it's barely longer its own type signature.

piop/src/lookup/booleanity.rs
```diff
@@ -263,6 +246,38 @@ where
...
+    pub fn prepare_affine_virtual_sumcheck_group<const D: usize>(
+        transcript: &mut impl Transcript,
+        trace_bin_poly: &[DenseMultilinearExtension<BinaryPoly<D>>],
+        affine_virtual_specs: &[AffineVirtualSpec],
+        num_vars: usize,
+        field_cfg: &C,
+    ) -> Result<(MultiDegreeSumcheckGroup<C>, BoolProverAncillary), BooleanityError<C::Element>>
+    {
+        if affine_virtual_specs.is_empty() {
+            return Err(BooleanityError::NoBinaryPolyColumns);
```
frozenspider, line 268: 1. This error does not match the condition being checked.
2. if there's actually no binary poly columns (`trace_bin_poly.len() == 0`), then `build_affine_virtual_bit_slice_mles` will panic due to `trace_bin_poly.first().expect(...)`

piop/src/lookup/booleanity.rs
```diff
@@ -1,11 +1,13 @@
...
-//! a bit $\in \\{0,1\\}$. The argument is structured as a single
-//! [`MultiDegreeSumcheckGroup`] of degree 3, batched alongside the existing
-//! CPR group with shared randomness, and emits bit-slice claims at the
+//! a bit $\in \\{0,1\\}$. The committed-witness path is structured as a
+//! [`MultiDegreeSumcheckGroup`] of degree 3, batched alongside the existing CPR
+//! group with shared randomness, and emits bit-slice claims at the
 //! multi-degree sumcheck output point $r^\star$ for the protocol layer's
 //! $\alpha'$ bridge into multipoint-eval (see "Soundness bridge" below).
+//! Affine virtual booleanity targets use a separate generic degree-3 group in
+//! the same sumcheck family.
```
frozenspider: Is that so though?

piop/src/lookup/booleanity.rs
```diff
@@ -685,6 +700,174 @@ where
...
+    C: SemiringConfig + ProjectPrimitiveIntegersWithConfig,
+{
+    if affine_virtual_specs.is_empty() {
+        return Vec::new();
+    }
+
+    let num_vars = trace_bin_poly
+        .first()
+        .expect("affine virtual specs need at least one binary source column")
+        .num_vars;
```
frozenspider: I think we should take `num_vars` as a parameter rather than taking it in this fashion

piop/src/lookup/booleanity.rs
```diff
@@ -919,6 +1105,182 @@ mod tests {
...
+            num_vars,
+            cfg,
+        );
+
+        if !expect_ok {
+            assert!(
+                matches!(verifier_anc, Err(BooleanityError::NonZeroClaimedSum { .. })),
+                "non-boolean affine residual should be rejected before sumcheck verify"
+            );
+            return;
```
frozenspider: So `expect_ok` covers only this specific case, which isn't great.
We should also test a malicious prover case, i.e. tamper the proof and make sure verifier fails.

uair/src/lib.rs
```diff
@@ -618,6 +618,17 @@ impl<Prime: Semiring> UairSignature<Prime> {
         &self.witness_cols
     }
 
+    /// Number of binary-polynomial booleanity targets.
+    ///
+    /// This is distinct from [`Self::witness_cols`]: affine virtuals are
+    /// Q-side booleanity targets, but they are not committed witness columns.
+    pub fn num_booleanity_bin_cols(&self) -> usize {
```
frozenspider: This is currently unused, please remove it and add it when the consumers would need it, right now it's hard to evaluate/review.

piop/src/lookup/booleanity.rs
```diff
@@ -685,6 +700,174 @@ where
...
+        for bit_idx in 0..D {
+            let mut evals: Vec<C::Element> = Vec::with_capacity(n_rows);
+            for row_idx in 0..n_rows {
+                let mut residual = ones.clone();
+                for (source_col, coeff, row_shift) in &terms {
+                    let shifted_row = add!(row_idx, *row_shift);
+                    if shifted_row >= n_rows {
+                        continue;
+                    }
+                    let bit = trace_bin_poly[*source_col].evaluations[shifted_row]
```
frozenspider: This is a big mismatch!
Here we're indexing inside `trace_bin_poly` (a **witness** slice!) using `source_col` that is [described](https://github.com/NethermindEth/zinc-plus/blob/8dbd6007008b2d10e95e73149ca2fd5b7d8e00f9/uair/src/lib.rs#L205-L207) as "using the same flat `binary_poly || arbitrary_poly || int` indexing as [`ShiftSpec`]", i.e. as described it's supposed to be using for indexing inside the full trace (which is public-then-witness).

uair/src/lib.rs
frozenspider: This is the comment I'm referring to.
Also please move those comments to their respective fields, it makes no sense for them to be at the struct level.

frozenspider (changes requested):

frozenspider:
Please update the PR description - e.g. `num_booleanity_bin_cols` is no longer here.

frozenspider (approved):
Thanks!

PR #225, opened 2026-07-19
Bind combined row before spot-check challenges

latifkasuli:
## What changed

`combined_row` was written to the proof stream without being absorbed into the Fiat-Shamir transcript. This change adds bound fixed-width read/write helpers and uses them on both prover and verifier before column-opening indices are derived.

## Why

The column indices are verifier challenges sampled after the prover supplies `combined_row`, so that row must be part of the transcript state used to derive them. This PR fixes that ordering issue; it does not rely on or include proof-of-work grinding.

The proof format length is unchanged, but downstream Fiat-Shamir challenges change.

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p zip-plus --lib` (86 passed)
- `cargo test -p zinc-protocol --lib` (30 passed)

The regression checks that different bound messages produce different subsequent challenges and that prover/verifier transcripts remain synchronized.

### DIFF
--- a/zip-plus/src/pcs/phase_prove.rs
+++ b/zip-plus/src/pcs/phase_prove.rs
@@ -241,7 +241,8 @@ impl<Zt: ZipTypes, Lc: LinearCode<Zt>> ZipPlus<Zt, Lc> {
             combined
         };
 
-        transcript.write_const_many(&combined_row)?;
+        // Bind the prover's row before deriving the spot-check columns.
+        transcript.write_const_many_bound(&combined_row)?;
         for _ in 0..Zt::NUM_COLUMN_OPENINGS {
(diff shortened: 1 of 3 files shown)


frozenspider:
Very good catch, thanks! However, I have a different vision of how this should be solved

latifkasuli:
Thanks! I’m happy to adapt it. What approach do you have in mind?  
The invariant I was targeting is that combined_row must affect the Fiat–Shamir transcript before squeeze_challenge_idx. I can preserve the regression test and rework the implementation around your preferred transcript design.

frozenspider:
Thank you @latifkasuli, but what I had in mind is a major architectural change so I did it in #231

PR #230, opened 2026-08-12
Bind unshifted affine virtuals to source columns

latifkasuli:
Goes towards #185, which blocks #91. Follow-up to #221 and #222.

This wires unshifted affine virtual Booleanity targets through the full protocol:
* Prover runs the affine Booleanity group alongside the committed binary-column group and records the two results as separate optional proof components.
* Verifier prepares and finalizes the two Booleanity proofs separately, with affine counts derived from the public UAIR signature.
* Public source endpoints are recomputed from the public trace, while witness source endpoints reuse the existing multipoint-evaluation/PCS bridge.
* The affine endpoint reconstructed from those sources is checked against the endpoint proved Boolean; a mismatch returns `AffineVirtualBridgeMismatch`.

Affine virtuals remain uncommitted $Q[X]$ Booleanity targets, so this adds no PCS openings or $F_q[X]$ padding. `UairSignature::with_affine_virtual_specs` rejects shifted affine terms during signature construction; supporting them requires binding row-shifted source projections rather than treating them as unshifted.

Tests cover the honest path and separate proof shapes, signature-derived length rejection, a Booleanity-preserving `b -> 1-b` source-binding attack, the public-only path with no witness binary columns, and signature-level shifted-term rejection.

The PR also sorts `transcript/Cargo.toml` dependencies so the repository's `cargo sort --workspace --check` pre-push step passes after #231.

Validation: `.githooks/pre-push`.

### DIFF
--- a/protocol/src/lib.rs
+++ b/protocol/src/lib.rs
@@ -11,13 +11,16 @@
 //! After the three compiler steps, the protocol continues with:
 //!
 //! - Combined CPR + Booleanity + Lookup multi-degree sumcheck (CPR group at
-//!   degree `max_deg+2`; optional booleanity group at degree 3 when the UAIR
-//!   has witness binary-poly columns; one lookup group per table type; shared
-//!   eval point `r*`)
+//!   degree `max_deg+2`; optional degree-3 booleanity groups for committed
+//!   witness binary columns and affine virtual targets; one lookup group per
+//!   table type; shared eval point `r*`)
 //! - $\alpha'$ bridge: squeeze a fresh challenge $\alpha'$ after the booleanity
 //!   `bit_slice_evals` are absorbed, and append one extra $\alpha'$-projected
 //!   MLE + up-eval to the multipoint-eval inputs per witness binary-poly column
 //!   (see `BooleanityChecker`)
+//! - Affine virtual bridge: collapse each unshifted virtual at the same
+//!   `alpha'` and compare it directly with the affine combination of its
+//!   public/committed source projections at `r*`
 //! - Multi-point evaluation sumcheck (combines up/down evals at `r*` into a
 //!   single evaluation point `r_0`)
 //! - Lift-and-project (unprojected MLE evaluations at `r_0`)
@@ -130,6 +133,9 @@ pub struct Proof<F> {
     /// has no witness binary-poly columns (the argument is omitted from
     /// the multi-degree sumcheck in that case).
     pub booleanity_proof: Option<BooleanityProof<F>>,
+    /// Affine-virtual booleanity argument proof. `None` when the UAIR has no
+    /// affine virtual specs.
+    pub affine_booleanity_proof: Option<BooleanityProof<F>>,
     /// Per-prime $F_{q_i}[X]$ ideal-check proofs, one per declared
     /// prime in [`zinc_uair::UairSignature::primes`], in the same order.
     /// Empty for UAIRs with $Q[X]$-only constraints.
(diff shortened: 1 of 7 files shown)


protocol/src/lib.rs
```diff
@@ -738,6 +756,115 @@ fn alpha_prime_bridge_up_evals<C: BaseFieldConfig, const D: usize>(
...
 }
 
+/// Reject shifted affine virtual terms until their source openings are bound
+/// through the row-shift multipoint-evaluation kernel.
+///
+/// An unshifted affine virtual can be reconstructed from its source columns'
+/// evaluations at the shared booleanity point. A shifted source cannot: its
+/// shifted MLE evaluation is not determined by the unshifted evaluation at the
+/// same point.
+fn ensure_affine_virtuals_are_unshifted<F, P>(
```
frozenspider: We already have one check inside `UairSignature::with_affine_virtual_specs`, I think this should be another one there instead of a standalone function.

protocol/src/prover.rs
```diff
@@ -1792,44 +1834,3 @@ fn commit_optionally<Zt: ZipTypes, Lc: LinearCode<Zt>>(
...
-    field_cfg: &C,
-) -> DenseMultilinearExtension<C::Element>
-where
-    C: BaseFieldConfig,
-{
-    debug_assert_eq!(alpha_powers.len(), D);
-    let zero = field_cfg.zero();
-
-    // Sequential row loop: per-row work is at most `D` conditional adds.
-    // The caller parallelizes the outer loop over witness binary-poly columns.
```
frozenspider, line 1816: The comment about why this loop should be sequential should be preserved

protocol/src/verifier.rs
```diff
@@ -869,11 +874,33 @@ where
...
+                    num_affine_virtuals,
+                    D,
+                    self.base.num_vars,
+                    &self.field_cfg,
+                )
+                .map_err(ProtocolError::Booleanity)?,
+            )
+        } else {
+            None
+        };
```
frozenspider: The code here is a literal copy-paste of the booleanity verifier check above. Stick to the DRY principle.
Same applies to `BooleanityChecker::<C>::finalize_verifier` calls below.

protocol/src/lib.rs
```diff
@@ -1245,6 +1373,206 @@ mod tests {
...
+        )
+        .expect_err("the affine source-binding bridge must reject the tamper");
+        assert!(matches!(
+            err,
+            ProtocolError::AffineVirtualBridgeMismatch { .. }
+        ));
+    }
+
+    #[test]
+    fn shifted_affine_virtuals_fail_closed_at_public_entries() {
```
frozenspider: Naming: Other tests in this file have `test_` prefix, this one doesn't

protocol/src/prover.rs
```diff
@@ -1102,19 +1125,37 @@ impl_with_type_bounds!(ProverEvalProjected
...
+            )
+            .map_err(ProtocolError::Booleanity)?;
+            bit_slice_evals.extend(proof.bit_slice_evals);
+        }
+        debug_assert!(md_iter.next().is_none());
+
+        let booleanity_proof = if bit_slice_evals.is_empty() {
             None
+        } else {
+            Some(BooleanityProof { bit_slice_evals })
```
frozenspider: So here on the prover side we concatenate booleanity proof with affine bool proof, but on the verifier side we split them back and use them separately. To my understanding, everything else in code treats them separately as well. To me that means they should not have been concatenated to begin with.

frozenspider (changes requested):

frozenspider:
@codex review

frozenspider:
@latifkasuli, please update the PR description - e.g. `ShiftedAffineVirtualUnsupported` no longer exists

frozenspider (approved):
Thanks!

